PR #19, opened 2019-04-10
Adding breakme.py

ervbrian:
Common exception examples


PythonCHB (approved):
nice work!

You went above and beyond -- but I promise the class will get more challenging as we go on :-)

PR #23, opened 2019-04-10
try_Sally_addReadme

momosally:
This is the first try


nairraghav:
Hi Sally, it looks like you created a file under students. You will want to create a directory under students and then a readme (txt is file) file inside that directory. Your structure will want to look like the following:
UWPCE-PythonCert-ClassRepos (folder) -> students (folder) -> Sally_Shen (folder) -> readme.txt (file)

PR #26, opened 2019-04-12
Assignment 1 submit

archwu:
Finished implement of break_me.py

### DIFF
--- a/students/ArchWu/Session1/break_me.py
+++ b/students/ArchWu/Session1/break_me.py
@@ -3,4 +3,19 @@ def name_error():
     pintx("lel")
 name_error()
 
-just a change to test
+def type_error():
+    x = '1' + 1
+    print(x)
+
+type_error()
+
+def syntax_error():
+    print()
+
+syntax_error()
+
+def attribute_error():
+    x = 1
+    x.append(0)
+
+attribute_error()
--- a/students/ArchWu/Session1/break_me.py
+++ b/students/ArchWu/Session1/break_me.py
@@ -1,21 +1,33 @@
+# deliberately implementing bugs to confuse python interpreter
+# name_error: exception when calling an undeclaired function/parameter
+# type_error: exception when adding two or more parameters which have different types
+# syntax_error: exception when missing or having extra punctuations. Violation of the python syntax.
+# attribute_error: exception when trying to access attributes of a certain parameter/object which does not have that attribute
+
+
+
 def name_error():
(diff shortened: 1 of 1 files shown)


students/ArchWu/Session1/break_me.py
```diff
@@ -0,0 +1,33 @@
...
+def type_error():
+    x = '1' + 1
+    # adding an int and a char triggers type error
+    print(x)
+
+type_error()
+
+def syntax_error():
+    # misuse of syntax
+    print()()
```
PythonCHB, line 24: this is an error -- but it's not a Syntax Error:

```
TypeError                                 Traceback (most recent call last)
<ipython-input-2-396ec0978f96> in <module>
----> 1 print()()

TypeError: 'NoneType' object is not callable
```

That's because when you call a function (like print), it can return anything, including another callable function. Here's a silly example:

```
In [3]: def get_print(): 
   ...:     return print 
   ...:                                                                         

In [4]: get_print()()                                                           


In [5]: get_print()("something to print")                                       
something to print
```
So you see that a_name()() is perfectly valid syntax.

However, the regular print function returns None, and None is not a "callable" or function -- that is, you can't put parentheses after it to run it -- soyou get the TypeError-- same as:

```
In [6]: None()                                                                  
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-6-061f79a46d11> in <module>
----> 1 None()

TypeError: 'NoneType' object is not callable
```

A Syntax Error is something that is not valid Python, regardless of any of the values of the variables. I'm sure you'll see them from typos, etc, Here's one:

```
In [9]: def a_fun: 
   ...:     return 5                                                            
  File "<ipython-input-9-e3dbe231ad8d>", line 1
    def a_fun:
             ^
SyntaxError: invalid syntax
```
That's a SyntaxError because there has to be parentheses after a function definition before the colon:

```
In [10]: def a_fun(): 
    ...:     return 5 
    ...: 
```

archwu: Hi Brian, thank you so much for pointing out this mistake. It is weird that when I was editing the python file, I was using the missing parentheses as the syntax error but I changed my mind to edit it to this because I feel extremely uncomfortable to see a non-symmetry parentheses in the code(especially my editor keep noticing me on that). Anyways, I should have run it before the final submission. Thanks again for correcting my mistake!

PR #27, opened 2019-04-13
Test submission Breakme

FredBallyns:
No need to review. 3 of 4 done during office hours.


students/FredB/session01/breakme.py
```diff
@@ -0,0 +1,24 @@
...
+
+#TypeError
+
+a=5
+b="5"
+c=a+b
+
+
+#SyntaxError
+"
```
PythonCHB: make sure to comment this out to be able to run the rest of the code.

students/FredB/session01/breakme.py
```diff
@@ -0,0 +1,19 @@
...
+def TypeError():
+a=5
+b="5"
+#c=a+b
+
+def SyntaxError():
+#"
+
+def AttributeError():
+List = 5
```
PythonCHB, line 16: did this compile and run?

you need to indent the code inside a function:

```
def attribute_error():
    List = 5
    LIst.append(4)
```
then you need to call the function to get it to run:

```
attribute_error()
```

And you don't want to call your function "AttributeError", as that will over wright the built in name for the Exception.

students/FredB/session01/warmups.py
```diff
@@ -0,0 +1,23 @@
...
+#Returns true if not weekday or vacation
+  return(not weekday or vacation)
+
+
+def pos_neg(a, b, negative):
+  if negative:
+    return (a < 0 and b < 0)
+  else:
+    return (a * b < 0)
+
```
PythonCHB, line 23: this is all looking good.

PythonCHB (commented):
I'm going to merge -- I'm sure you'll see those errors soon enough :-)

PR #28, opened 2019-04-13
break_me.py

trayffs:
break_me assignment file.

I simply had to comment on the theory of syntax error as I couldn't find an example that wasn't fatal.

### DIFF
--- a/students/Tyler Ray/session1/break_me.py
+++ b/students/Tyler Ray/session1/break_me.py
@@ -1,0 +1,44 @@
+"""
+breakme
+"""
+
+def name_error():
+    print("the name error function")
+    printd("something") #spelled print wrong--aka type or reference something that doesn't exist
+
+def type_error():
+    print("Please input an integer to solve the following equation: X + 1")
+    vx = input() # Does not require an int input therefore the following won't work:
+    print(vx + 1) # Can't add an integer and a string
+
+def syntax_error(): # As just one example of a syntax error: If i removed the ':' from syntax_error() then I'd get a fatal syntax error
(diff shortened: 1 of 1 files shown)


students/Tyler Ray/session1/break_me.py
```diff
@@ -0,0 +1,44 @@
...
+def name_error():
+    print("the name error function")
+    printd("something") #spelled print wrong--aka type or reference something that doesn't exist
+
+def type_error():
+    print("Please input an integer to solve the following equation: X + 1")
+    vx = input() # Does not require an int input therefore the following won't work:
+    print(vx + 1) # Can't add an integer and a string
+
+def syntax_error(): # As just one example of a syntax error: If i removed the ':' from syntax_error() then I'd get a fatal syntax error
```
PythonCHB, line 14: yes, this was the idea -- you can't recover from a syntax error. The file won't compile, and none of the code in the file will run.

The idea was to get used to that :-)

PythonCHB:
Nicely done -- you are ahead of the curve here -- I promise the class will get more challenging.

PythonCHB:
one more note:

While every OS allows spaces in filenames, it can get awkward when using the command line.

If you want to change teh name of your dir, you can do:

`git mv "Tyler Ray" TylerRay

trayffs:
I figured out I can just use "quotes" to combine lengthy, space separated
directories like "Tyler Ray".

PythonCHB:
yup -- that's the trick -- but it does get tiresome sometimes.

No problem for me this way though.

PR #29, opened 2019-04-14
warmup.py

trayffs:
This is the "monkey trouble" warm up from CodingBat. The instructions are in the .py file.

### DIFF
--- a/students/Tyler Ray/session1/warmup.py
+++ b/students/Tyler Ray/session1/warmup.py
@@ -1,0 +1,11 @@
+'''
+Given an int n, return True if it is within 10 of 100 or 200. Note: abs(num) computes the absolute value of a number.
+
+near_hundred(93) → True
+near_hundred(90) → True
+near_hundred(89) → False
+
+'''
+
+def near_hundred(n):
+  return ( n in range(90,111) or n in range(190,211))


students/Tyler Ray/session1/warmup.py
```diff
@@ -0,0 +1,11 @@
...
+Given an int n, return True if it is within 10 of 100 or 200. Note: abs(num) computes the absolute value of a number.
+
+near_hundred(93) → True
+near_hundred(90) → True
+near_hundred(89) → False
+
+'''
+
+def near_hundred(n):
+  return ( n in range(90,111) or n in range(190,211))
```
PythonCHB, line 11: this will sure work, but would be pretty inefficient if you were working with larger ranges of numbers.

Also, it would fail for non-integers -- granted, the problem did specify an int, but particular with a dynamically typed language like Python, where floats and integers can mostly be used interchangeably, it's better not to assume.

Can you do it with a computation, rather than checking for all the values?

trayffs:
I agree and after seeing the suggested solution my code felt rather clunky
or inflexible. I hope that with more experience and training of my mind in
the logic of Python I'll develope more finesse with the code.

PR #30, opened 2019-04-14
Lesson01 and lesson02

gjkim44:
I added the files early for lesson01 and lesson02.  I started Assignment 02, fizzbuzz and submitted it

### DIFF
--- a/students/GKim/lesson01/maketags.py
+++ b/students/GKim/lesson01/maketags.py
@@ -1,0 +1,18 @@
+# The web is built with HTML strings like "<i>Yay</i>" which draws Yay as italic text. 
+# In this example, the "i" tag makes <i> and </i> which surround the word "Yay". 
+# Given tag and word strings, create the HTML string with tags around the word, e.g. "<i>Yay</i>".
+
+
+# make_tags('i', 'Yay') → '<i>Yay</i>'
+# make_tags('i', 'Hello') → '<i>Hello</i>'
+# make_tags('cite', 'Yay') → '<cite>Yay</cite>'
+
+def make_tags(tag, word):
+    front = "<" + tag +">"
+    end = "</" + tag + ">"
+    mtag = front + word + end
+    return mtag
+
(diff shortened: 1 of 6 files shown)


nairraghav:
Great job George! As a side note, It looks like you created two directors for session (lesson) 1. Moving forward, I think you should be fine with creating one since it looks like it's just a duplicate.

students/GKim/lesson01/maketags.py
```diff
@@ -0,0 +1,18 @@
+# The web is built with HTML strings like "<i>Yay</i>" which draws Yay as italic text. 
+# In this example, the "i" tag makes <i> and </i> which surround the word "Yay". 
+# Given tag and word strings, create the HTML string with tags around the word, e.g. "<i>Yay</i>".
+
+
+# make_tags('i', 'Yay') → '<i>Yay</i>'
+# make_tags('i', 'Hello') → '<i>Hello</i>'
+# make_tags('cite', 'Yay') → '<cite>Yay</cite>'
+
+def make_tags(tag, word):
```
PythonCHB, line 10: note for the future -- we'll learn about string formatting in a couple weeks -- you may want to come back to this then.

gjkim44: is anything wrong with it?

gjkim44: I guess I could have done this
def make_tags(tag, word):
   return "{}{}{}{}{}{}{}".format("<",tag,">",word,"</",tag,">")

PythonCHB: actually, what you want to do with string formatting, is put the text in teh string:

"<{}>{}</{}>".format(tag, word, tag)

(untested ...)

gjkim44: doah!  Ah, ok.  It's all coming back.  Getting my mind back into python mode after about 3 weeks off.

PR #32, opened 2019-04-14
lesson1 break_me

momosally:
To trigger each error you need to comment out the previous Trigger...(). If not trigger syntax error, comment out import.


nairraghav:
Nice job!

PR #34, opened 2019-04-15
completed warmup1

momosally:


students/Sally_Shen/lesson1/warmup1.py
```diff
@@ -0,0 +1,78 @@
...
+        new_str = str[:n] + str[(n + 1):]
+        return new_str
+    else:
+        return str
+
+
+
+def front_back(str):
+    if len(str) > 1:
+        new_str = str[-1] + str[1:len(str)-1] + str[0]
```
PythonCHB, line 65: we'll cover "slicing" in a week or so -- this could be done a bit easier:

`new_str = str[-1] + str[1:l-1] + str[0]`

PythonCHB:
nice job!

PythonCHB (approved):

PR #35, opened 2019-04-15
adding lesson01 break me file

BBrumleve:
Uploading my code for lesson 01 creating exceptions.


lesson01/break_me.py
```diff
@@ -0,0 +1,31 @@
...
+#    return(date)
+#print(type_error())
+#
+#
+## I recall it was stated in class SyntaxError would not
+## be used.  Well here's my hack at it anyway.
+#
+## SyntaxError
+#def syntax_error(x,y):
+#    sum = x + 3y
```
PythonCHB: yup!

lesson01/break_me.py
```diff
@@ -0,0 +1,31 @@
...
+#    return(sum)
+#print(syntax_error(2,3))
+#
+#
+## AttributeError
+#def attribute_error():
+#    age = 8
+#    age = age.append(3) # integers cannot be appended. Python appends objects.
```
PythonCHB: you can append an integer, but you have to append it to a list...

lesson01/sum_double.py
```diff
@@ -0,0 +1,19 @@
...
+#
+# sum_double(1, 2) → 3
+# sum_double(3, 2) → 5
+# sum_double(2, 2) → 8
+#------------------------------------------
+
+def sum_double(a,b):
+    if a != b:
+        return(a+b)
+    elif a==b:
```
PythonCHB: wouldn't this always be true if a!=b failed?

PythonCHB (approved):

PythonCHB:
Your work is fine, but you've put the lesson01 dir in the main repo directory. 

You need to put all you rstuff in your own dir in students. You should be able to do that with:

```
git mv lesson01 students/BrianB/
git commit -a -m "moved my lesson01 dir into my dir"
git push
```

If you are BrainB, anyway :-)

BBrumleve:
BrianB == me :-). This is the correct person.

Do I need to submit a new pull request? Or do i Need to do anything else
from within gitHub itself? I’ve moved the folder location by following the
commands you showed in your email.

Thanks,
BrianB

nairraghav:
Looks good Brian! As Chris has noted, you would not need to specify the elif in your sum_double since it would be guaranteed (otherwise it would be caught in your if statement).

PR #36, opened 2019-04-15
lesson 1_task 2 - sample code

jchando1:


students/JasneetChandok/Lesson01/Task02.py
```diff
@@ -0,0 +1,33 @@
...
+'''
+#!/usr/bin/env python3.6
+
+# Warmup -1 > sleep_in
+def sleep_in(weekday, vacation):
+  if not weekday or vacation:
+    return True
+  else:
+    return False
+  # This can be shortened to: return(not weekday or vacation)
```
PythonCHB, line 15: indeed it can -- so why not shorten it? :-)

students/JasneetChandok/Lesson01/Task02.py
```diff
@@ -0,0 +1,33 @@
...
+  return False
+
+# Warmup -1 > sum_double
+def sum_double(a, b):
+  # Store the sum in a local variable
+  sum = a + b
+  
+  # Double it if a and b are the same
+  if a == b:
+    sum = sum * 2
```
PythonCHB, line 32: you could use sum *= 2 -- though this is fine.

PythonCHB (approved):

PR #37, opened 2019-04-15
Python Push-ups Part 1

Charlie-England:


Charlie-England:
Adding Lesson 1 warmup problem solution and adding error testing

students/charlie_england/session01/break_me.py
```diff
@@ -0,0 +1,25 @@
...
+    #parsing errors
+    #fr x in range(1,10):
+        #print(x)
+    pass
+
+def attribute_error():
+    #trying to use an attribute (function) of an object when the function does not exist
+    integer = 8
+    integer.append(1) #cannot append to an integeter, can append to a list
+
```
PythonCHB, line 25: perfect.

students/charlie_england/session01/lesson1.py
```diff
@@ -0,0 +1,7 @@
+def sleep_in(weekday, vacation):
+    if weekday and not vacation:
+        return False
+    if not weekday:
```
PythonCHB, line 4: if an "if" and "else" both do the same thing -- do you need it at all?

PythonCHB (approved):

students/charlie_england/session01/lesson1.py
Charlie-England: Yep, you don't need both. That was a silly mistake. Also, I think this was an older solution that I did on that account. this is a better one.

def sleep_in(weekday, vacation):
  if weekday and not vacation:
    return False
  else:
    return True

PR #39, opened 2019-04-15
Add python pushups part 1.

JoeNunnelley:
Session01 Work added:  Pushups -> Went through a fair number of these excercises.  
Session02 Work In Progress added:  The print_grid files is still WIP.

### DIFF
--- a/students/JoeNunnelley/session01/break_me.py
+++ b/students/JoeNunnelley/session01/break_me.py
@@ -1,30 +1,42 @@
 #! /usr/bin/env python3
 
-""" break me : output error handling """
-"""
-Errors:
-  NameError, TypeError, SyntaxError, AttributeError
+""" break me : output error handling
+    Errors:
+    NameError, TypeError, SyntaxError, AttributeError
 """
 
 def name_error():
     print("the name error function")
-    printx("hello")
-
-name_error()
-
+    try:
+        printx("hello")
+    except NameError:
+        print('name exception caught')
 
 def type_error():
-    print("the type error function")
-
-type_error
+    print('the type error function')
+    try:
+       mylist = [4,5]
+       print("2" + mylist)
+    except TypeError:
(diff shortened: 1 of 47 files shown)


PythonCHB:
print_grid is for Lesson2 -- so we are expecting you to work on it after this weeks class :-)

No harm in getting ahead though.

students/JoeNunnelley/session01/break_me.py
```diff
@@ -0,0 +1,42 @@
...
+       print("uncomment line below to see an actual SyntaxError")
+       #4.split()
+    except SyntaxError:
+        print('syntax error caught')
+
+
+name_error()
+type_error()
+attribute_error()
+syntax_error()
```
PythonCHB, line 42: We weren't expecting you to catch the errors -- but nice job!

Did you notice that you never caught the syntax error though? that one is raised when teh code is parsed -- so it can't actually run and be caught by a try: except block.

students/JoeNunnelley/session01/warmups.py
```diff
@@ -0,0 +1,263 @@
...
+#29
+def double_char(str):
+    new_str = ''
+    for i in str:
+        new_str += i * 2
+
+    return new_str
+
+#30
+def count_code(str):
```
PythonCHB, line 254: did coding bat use the parameter name "str" ? Careful that's a python built it for the string object -- you don't want to overwrite it.

students/JoeNunnelley/session01/warmups.py
```diff
@@ -0,0 +1,263 @@
...
+    match = 0
+    i = 0
+    while i < len(str) - 3:
```
PythonCHB, line 257: could you do this with a for loop and range?

PythonCHB (approved):
nice job! those coding bat puzzles are kind of fun, yes?

PR #42, opened 2019-04-16
adding part 1 pushups

roviram:


students/roviram/lesson01/Lesson_01_Part1_Python_Pushups.py
```diff
@@ -0,0 +1,13 @@
+#---Dev Script---#
```
PythonCHB, line 1: this is in the "Lesson_01" dir -- so need for such a long name .. :-)

PythonCHB (approved):

PythonCHB (approved):

PR #43, opened 2019-04-17
grid print exercise, test passed

kanmenlaohan:


students/Lincoln_Zhang/Session2/grid_print.py
```diff
@@ -0,0 +1,53 @@
...
+#|         |         |
+#|         |         |
+#|         |         |
+#+ - - - - + - - - - +
+
+def print_plus():
+	print ("+",end="")
+
+def print_minus():
+	print(" - - - - ",end="")
```
nairraghav: The parameter we are passing into grid_print should determine the size. It should not be a hardcoded amount of 4 minuses. It looks like you are calculating the "size" of the grid to be the length from top to down. Consider it to be the full length top down and across so a size of 3 should be 3x3.

students/Lincoln_Zhang/Session2/grid_print.py
```diff
@@ -0,0 +1,53 @@
...
+
+def print_pipe():
+	print("|",end="")
+
+def print_newline():
+	print("\n")
+
+def grid_print(r):
+	for r in range(r):
+		if r%5 == 0:
```
nairraghav: This print only prints the top half of the grid, you will want to have a grid of 4 squares at the end.

nairraghav (changes requested):

kanmenlaohan:
assignment 2 from Lincoln

students/ArchWu/Session2/series.py
```diff
@@ -0,0 +1,84 @@
...
+    This function should generalize the fibonacci() and the lucas(),
+    so that this function works for any first two numbers for a sum series.
+    Once generalized that way, sum_series(n, 0, 1) should be equivalent to fibonacci(n).
+    And sum_series(n, 2, 1) should be equivalent to lucas(n).
+    """
+    if n0 == 0 and n1 == 1:
+        return fib2(n)
+    elif n0 == 2 and n1 == 1:
+        return lucas(n)
+    else:
```
nairraghav: Nice job! This else generalized logic should also be able to handle n0,n1 = 0,1 and n0,n1 = 2,1

students/Lincoln_Zhang/Session2/series.py
```diff
@@ -0,0 +1,23 @@
...
+
+	l = [num_1,num_2]
+	if num_1 == 0 and num_2 ==1:
+		print("A Fibonacci series is generated.")
+	elif num_1 == 2 and num_2 == 1:
+		print("A Lucas series is generated.")
+	else:
+		print("A series is generated.")
+	while len(l) <= n:
+		new_num = l[-1] + l[-2]
```
nairraghav, line 18: Interesting method 😄 Nice job, this generalizes the logic for all types of series, good job!

students/Lincoln_Zhang/Session2/grid_print.py
```diff
@@ -0,0 +1,60 @@
...
+# 	r = horizontal_line(boxsize)
+# 	c = vertical_line(boxsize)
+# 	for _ in range(gridsize+1):
+# 		if _%boxsize == 0:
+# 			print ("+ " + r*gridsize)
+# 		else:
+# 			for _ in range(boxsize):
+# 				print ("| " + c*gridsize)
+
+def grid_print_2(boxsize, gridsize):
```
nairraghav, line 49: Nice, grid_print_2 looks great!

students/Lincoln_Zhang/Session2/grid_print.py
```diff
@@ -0,0 +1,60 @@
...
+boxsize = int(input("Please input box size: "))
+gridsize = int(input("Please input grid size: "))
+
+def horizontal_line(boxsize):
+	return "- " * boxsize + "+ "
+
+def vertical_line(boxsize):
+	return "  " * boxsize + "| " 
+
+# def grid_print_1(boxsize, gridsize):
```
nairraghav: We will still want to get grid_print_1 working. The initial print_grid should always create a grid of 4 cells with the total height/length being the parameter we give it. For example: print_grid(9) should return a 2x2 grid where the entire length/height (2 cells worth) is 9. Hint: The middle '+' will handle it such that the each individual cells are of 4 length each.

kanmenlaohan: A commit for this feedback is pushed

students/Lincoln_Zhang/Session2/grid_print.py
```diff
@@ -21,22 +21,28 @@
...
-# 	r = horizontal_line(boxsize)
-# 	c = vertical_line(boxsize)
-# 	for _ in range(gridsize+1):
-# 		if _%boxsize == 0:
-# 			print ("+ " + r*gridsize)
-# 		else:
-# 			for _ in range(boxsize):
-# 				print ("| " + c*gridsize)
+#a scale function for a 4X4 grid
+def grid_print_1(n):
```
nairraghav: I'm not too sure what this is doing. We will not need to sanitize any input to make sure it is modulo 4. 

Refer to the first example shown: https://uwpce-pythoncert.github.io/PythonCertDevel/exercises/grid_printer.html

At the very top, we print a 2x2 grid with grid_print(9). The 9 here represents the total length of 1 side. If you count the number of '-' characters (and the '+') in the middle, we see a total of 9. This is what the parameter is responsible for. Similarly, follow the grid_print(3) and grid_print(15) examples, they still create a 4x4 grid but the sides are of length 3 and 15 respectively.

kanmenlaohan: Thanks I see what you are saying. check out this revision please.

students/Lincoln_Zhang/Session2/fizzbuzz.py
```diff
@@ -0,0 +1,57 @@
+#!#!/usr/bin/env python3
+# Write a program that prints the numbers from 1 to 100 inclusive.
+# But for multiples of three print “Fizz” instead of the number.
+# For the multiples of five print “Buzz” instead of the number.
+# For numbers which are multiples of both three and five print “FizzBuzz” instead.
+
```
PythonCHB, line 6: looks like you've got <tabs> in there. You REALLY want to use only four spaces for indents.

Make sure your editor is configured to put in 4 spaces when you hit the <tab> key.

students/Lincoln_Zhang/Session2/fizzbuzz.py
```diff
@@ -0,0 +1,57 @@
...
+	if not n%15:
+		return "FizzBuzz"
+	else:
+		return n
+
+
+def Fizz_Buzz_filter(n):
+	l = [return_Fizz(n),return_Buzz(n),return_FizzBuzz(n)]
+	if "Fizz" in l:
+		return "Fizz"
```
PythonCHB, line 35: you have the functions return the word, then the hard-coding the word here....

students/Lincoln_Zhang/Session2/fizzbuzz.py
```diff
@@ -0,0 +1,57 @@
...
+	elif "Buzz" in l:
+		return "Buzz"
+	elif "FizzBuzz" in l:
+		return "FizzBuzz"
```
PythonCHB, line 39: if you check for FizzBuzz firt, you wouldn't need to do the full check for both numbers.

students/Lincoln_Zhang/Session2/grid_print.py
```diff
@@ -0,0 +1,67 @@
...
+|         |         |
+|         |         |
+|         |         |
++ - - - - + - - - - +
+|         |         |
+|         |         |
+|         |         |
+|         |         |
++ - - - - + - - - - +
+""")
```
PythonCHB, line 20: like the quick and dirty version :-)

students/Lincoln_Zhang/Session2/grid_print.py
```diff
@@ -0,0 +1,67 @@
...
+n = int(input("Please input a positive integer: ")) 
+
+def horizontal_line(boxsize):
+	return "- " * boxsize + "+ "
+
+def vertical_line(boxsize):
+	return "  " * boxsize + "| " 
+
+
+def grid_print_1(n):
```
PythonCHB, line 33: feels a little complicated, but nicely done!

students/Lincoln_Zhang/Session2/series.py
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env python3
+
+
+
+num_1 = int(input("Please input the 1st number: "))
+num_2 = int(input("Please input the 2nd number: "))
+
+def fibonacci(n):
+
+	l = [num_1,num_2]
```
PythonCHB, line 10: would be a bit better (and the assignment :-) ) if num_1 and num_2 were passed in to the function.

But you've got the logic down.

PythonCHB (approved):
nice work!

PR #47, opened 2019-04-17
session 02 homework

JoeNunnelley:


students/JoeNunnelley/session02/series.py
```diff
@@ -0,0 +1,116 @@
...
+
+    if not first:
+        """Return the fibonacci number for the value provided"""
+        if n > 1:
+            return fibonacci(n - 1) + fibonacci(n - 2)
+        if n == 1:
+            return second
+        elif n <= 0:
+            return first
+    else:
```
nairraghav, line 58: Hey Joe, it seems I spoke too early for this submission. You handle the Fibonacci and Lucas cases but you will want to have a generalized solution that can handle any n0 and n1. Look into a solution that will satisfy any choice of n0 and n1 (this way it will also handle fibonacci and lucas). Please submit another PR when you have the solution.

Let me know if you need me to clarify anything, thanks!

PR #49, opened 2019-04-17
Added grid1.py

RimleeTalukdar44:
Added grid.py


nairraghav:
Great job, don't forget to add the initial print_grid which only takes 1 argument.

PR #50, opened 2019-04-17
Adds grid_printer and breakme from jsward

jamessward:


students/jsward/session01/breakme.py
```diff
@@ -0,0 +1,7 @@
+# Break things
+
+def name_error():
+    #print("blah{}".format(undefined_variable))
+    pass
+
```
nairraghav: This is a great start! Don't forget to add the other errors: TypeError, AttributeError, SyntaxError

students/jsward/session02/grid_printer.py
```diff
@@ -0,0 +1,24 @@
...
+        if i % (grid_size +1) == 0 or i == (total_size -1):
+            grid_square = "+{}".format('-' * grid_size)
+            print("{}+".format(grid_square * grid_qty))
+        else:
+            grid_square = "|{}".format(' ' * grid_size)
+            print("{}|".format(grid_square * grid_qty))
+        i += 1
+
+
+grid_size = input("Cell size: ")
```
nairraghav: This is great, but it won't run if you try it. Don't forget to change your input from a String to an Int 😄

jamessward: so oddly, this works from the terminal in pycharm, but I did encounter the type error when I run the code with the debugger.  Added type conversion to fix this issue.

PythonCHB: That's suspicious -- is there a chance that PyCharm is using Python 2 when you run it?

Python 2 had a different `input()` function.

students/jsward/session02/fizz_buzz.py
```diff
@@ -0,0 +1,20 @@
...
+#fizz_buzz.py
+# write 1 to 100 inclusive
+# for multiples of 3 print "Fizz"
+# for multiples of 5 print "Buzz"
+
+
+def fb():
+    for i in range(100):
+        fb_string = ""
+        if i % 3 == 0:
```
PythonCHB, line 11: very nice! good catch on not having to check for %3  or %5 more than once!

students/jsward/session02/fizz_buzz.py
```diff
@@ -0,0 +1,20 @@
...
+            fb_string += 'Fizz'
+        if i % 5 == 0:
+            fb_string += 'Buzz'
+        if len(fb_string) < 1:
```
PythonCHB, line 15: this can usut be:

```
if not fb_string
```

an empty string is False -- there's a video about that next week :-)

students/jsward/session02/series.py
```diff
@@ -0,0 +1,79 @@
...
+def fibonacci(nums_to_compute):
+    # First solution
+    """
+    Print the Fibronacci number series.
+    :param nums_to_compute:
+    :return: none
+    """
+    results = [0, 1]
+    while len(results) <= nums_to_compute:
+        results.append(sum(results[-2:]))
```
PythonCHB, line 12: this is pretty clever -- nice!

students/jsward/session02/series.py
```diff
@@ -0,0 +1,79 @@
...
+    :param nth: number to return from within the series
+    :param n1: 1st number from which to start the series
+    :param n2: 2nd number from which to start the series
+    :return: nth number from the series
+    """
+    results = [n1, n2]
+    while len(results) <= nth:
+        results.append(sum(results[-2:]))
+    return results[nth]
+
```
PythonCHB, line 48: you do want to be careful about storing the whole series -- computer memory is big these days, but it's not a good habit to store a bunch of stuff you will never use.

students/jsward/session02/fizz_buzz.py
```diff
@@ -0,0 +1,20 @@
+
+#fizz_buzz.py
+# write 1 to 100 inclusive
+# for multiples of 3 print "Fizz"
+# for multiples of 5 print "Buzz"
+
+
+def fb():
```
nairraghav, line 8: Interesting way to handle it, nice!

students/jsward/session01/break_me.py
```diff
@@ -0,0 +1,24 @@
+# Break things
```
nairraghav, line 1: Nice job!

PythonCHB (approved):
really nice work!

students/jsward/session02/grid_printer.py
```diff
@@ -0,0 +1,42 @@
+
+def print_grid(grid_size, grid_qty):
```
nairraghav, line 2: I believe this is just a copy of your original print_grid2. You can just use the other function for the initial, the only thing you would want to do is hardcode the cellsize to be 2. One way to handle that is to make the default value 2:
def print_grid2(grid_size, grid_qty=2):

Nice job, overall!

PR #51, opened 2019-04-17
grid file in class

colephalen:


students/ColeP/.idea/vcs.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
nairraghav: When committing files, you do not need to commit changes that come from your IDE. Usually this can make things messy, I would recommend taking this out (and not including it in future Pull Requests)

students/ColeP/Session02/fib.py
```diff
@@ -0,0 +1,11 @@
+def fib(n):
+    '''compute the nth fibonacci number'''
+    if n == 0:
+        return 0
+    if n == 1:
+        return 1
+    if n > 1:
+        return n + fib(n-1)
+
```
nairraghav, line 9: Great start! Don't forget to add the lucas series and the sum_series functions.

students/ColeP/Session02/grid_print.py
```diff
@@ -0,0 +1,22 @@
+# horizontal = ('+ ' + ('- ' * 4)) * 2 + '+\n'
```
nairraghav, line 1: Very interesting way to do this, nice job!

PR #52, opened 2019-04-17
session 02 homework updates 

JoeNunnelley:
- add author comments
- add usage function to print_grid.py
- add missing flag handler logic
- add in-class created fast_fib function as an example of non-recursive fibonacci which is much more performant.


nairraghav:
Fantastic job Joe!

PR #53, opened 2019-04-18
Completed Grid Buzz Ex.

jchando1:


students/JasneetChandok/Lesson02/FizzBuzz.py
```diff
@@ -0,0 +1,21 @@
+'''
+Lesson: 02
+Excercise: Fizz Buzz
+Student Name: Jasneet Chandok
+Last Update Date: 04/18/19
+'''
+#!/usr/bin/env python3.6
```
PythonCHB, line 7: if you have a line like this, it should go at the very top of the file, it will only do any good on the first line.

students/JasneetChandok/Lesson02/FizzBuzz.py
```diff
@@ -0,0 +1,21 @@
...
+        if num % 3 == 0 and num % 5 == 0:
+            print ('FizzBuzz')
+        elif num % 5 == 0:
+            print ('Buzz')
+        elif num % 3 == 0:
+            print('Fizz')
+        else:
+        	print(num)
+
+fizz_buzz()
```
PythonCHB, line 21: solid solution.

PR #54, opened 2019-04-19
Lesson 02 finished

gjkim44:
I finished lesson02 assignments.  I got help from you tube for grid printer.   I also used the series_template to finish the fibonacci series exercise.


students/GKim/lesson02/series.py
```diff
@@ -0,0 +1,66 @@
...
+    so that this function works for any first two numbers for a sum series.
+    Once generalized that way, sum_series(n, 0, 1) should be equivalent to fibonacci(n).
+    And sum_series(n, 2, 1) should be equivalent to lucas(n).
+    """
+    if n0 == None and n1 == None:
+        return fibonacci(n)
+
+    elif n0 == 2 and n1 == 1:
+        return lucas(n)
+
```
nairraghav: This is very close! Remember that you will also have to handle series that don't start with 2,1 or no optional parameters. If n0 = 5 and n1 = 2 (or almost any other permutation), this sum_series won't do anything.

Hint: Your fibonacci and lucas functions are extremely similar. Think about how you can use one general function to do both of their jobs

students/GKim/lesson02/grid_printer.py
```diff
@@ -0,0 +1,45 @@
...
+    """
+    prints the "columns" " | " in the grid. just like the row first section is 
+    already created by the input given from user.  The colums print across 
+    """
+    print("|", end = " ")
+    for _ in range(c - 1):
+        print("        |", end = " ")
+    print("        |")
+
+def my_grid(c,r):
```
nairraghav: This is close but it doesn't achieve what we are aiming for. This prints out the rows/columns but we are looking for a grid where the rows/columns are the same. We are looking to grind a grid (with same rows/columns) but the **size** of each individual cell is another factor.

nairraghav:
Series looks good! Please take a look at the Grid Printer comments :)

students/GKim/lesson02/grid_printer.py
gjkim44: Ok I will make changes

students/GKim/lesson02/series.py
```diff
@@ -33,12 +33,13 @@ def sum_series(n, n0=None, n1=None):
     Once generalized that way, sum_series(n, 0, 1) should be equivalent to fibonacci(n).
     And sum_series(n, 2, 1) should be equivalent to lucas(n).
     """
-    if n0 == None and n1 == None:
-        return fibonacci(n)
-
-    elif n0 == 2 and n1 == 1:
-        return lucas(n)
-
+    if n == 0:
```
nairraghav, line 36: Nice job generalizing the function!

students/GKim/lesson02/grid_printer.py
```diff
@@ -0,0 +1,48 @@
...
+    prints the "columns" " | " in the grid. just like the row first section is 
+    already created by the input given from user.  The colums print across 
+    """
+    for _ in range(a-1):
+        print("|", end = "")
+        print("   " * ((c-1)//2), end = "")
+    print("|")
+    
+
+def my_grid(c,a =3):
```
nairraghav: This looks to be printing properly when given 1 parameter, ex: my_grid(5), my_grid(7). 

It doesn't properly handle the case when we pass 2 parameters though. The first parameter determines the nxn grid size and the second parameter determines the size of each cell. So for example, print_grid(5,3) should return a 5x5 grid where each cell is 3x3 (3 '-' across and 3 '|' long )

students/GKim/lesson02/series.py
nairraghav: This looks good, nice job!

gjkim44:
I shortened up the grid printer file to have 3 total functions!

students/GKim/lesson02/fizz_buzz.py
```diff
@@ -1,6 +1,4 @@
 
-
-
 for x in range(1,101):
     if x % 15 == 0:
```
PythonCHB, line 3: nice catch with the % 15 :-)

students/GKim/lesson02/grid_printer.py
```diff
@@ -0,0 +1,118 @@
+
+"""
+Shortened version
+"""
+
+def print_symbol(c, a =2,s1="",s2=""):
```
PythonCHB, line 6: be consistent with your spacing, ideally following PEP8:

```
def print_symbol(c, a=2, s1="", s2=""):
```
make it a little more readable ...

students/GKim/lesson02/grid_printer.py
```diff
@@ -0,0 +1,118 @@
...
+    """
+    function to print + and - using a 
+    for loop .  Using input from user the range is -1 because
+    from the first intial print the  section of the column is already created
+    this all prints accross
+    """
+    if c % 2 == 0:
+        x = int(c)/2
+    elif c % 2 == 1:
+        x = int(c-1)/2
```
PythonCHB, line 16: you can replace all this logic with "floor division":

```
x = c // 2
```

students/GKim/lesson02/grid_printer.py
```diff
@@ -0,0 +1,118 @@
...
+    for _ in range(a):
+        print("{}".format(s1), end = "")
```
PythonCHB, line 18: This: `"{}".format(s1)` is EXACTLY the same as simply `s1` :-)

students/GKim/lesson02/series.py
```diff
@@ -0,0 +1,73 @@
...
+    return b
+    
+
+
+def lucas(n):
+    """ compute the nth Lucas number """
+    a, b = 2, 1
+    if n == 0:
+        return a
+    elif n == 1:
```
PythonCHB, line 16: did you try without this check? turns out you don't need it ... can you see why?

students/GKim/lesson03/list_lab.py
```diff
@@ -0,0 +1,96 @@
+
```
PythonCHB, line 1: now you are getting ahead of yourself -- what will you do in class?

PythonCHB (approved):
nice work!

I haven't reviewed the list lab and slicing yet -- be sure to ask for a review of those if you want them when you do your next PR.

PR #55, opened 2019-04-19
Adding session02 code

ervbrian:
Added GridPrinter.py for displaying ascii grid patterns. 
Added series.py for calculating Fibonacci, Lucas and custom summation values.

### DIFF
--- a/students/BrianE/session02/GridPrinter.py
+++ b/students/BrianE/session02/GridPrinter.py
@@ -34,6 +34,7 @@ def grid(box_size=2, grid_size=2):
 
 def main():
     print('box_size=5 (single argument, default 2x2 grid):', grid(5), sep='\n')
+    print()
     print('box_size=1, grid_size=2:', grid(box_size=1, grid_size=2), sep='\n')
     print()
     print('box_size=2, grid_size=4:', grid(box_size=2, grid_size=4), sep='\n')
--- a/students/BrianE/session02/GridPrinter.py
+++ b/students/BrianE/session02/GridPrinter.py
@@ -31,10 +31,11 @@ def grid(box_size=2, grid_size=2):
     output += row_divider
     return output
 
+
 def main():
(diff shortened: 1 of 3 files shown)


nairraghav:
Great job Brian! It looks like you're missing the initial grid printer function, otherwise the rest looks great.

students/BrianE/session02/GridPrinter.py
```diff
@@ -0,0 +1,43 @@
+"""
+GridPrinter.py
+
+Prints ascii grid based on desired box size and grid size.
+"""
+
+
```
nairraghav, line 7: Don't forget to implement the initial grid which only takes 1 parameter.

students/BrianE/session02/series.py
```diff
@@ -0,0 +1,87 @@
...
+    assert lucas(6) == 18
+    assert lucas(7) == 29
+
+    # test if sum_series matched fibonacci
+    assert sum_series(5) == fibonacci(5)
+
+    # test if sum_series matched lucas
+    assert sum_series(5, 2, 1) == lucas(5)
+    print("All assertions pass")
+
```
nairraghav, line 94: Just a side note, you may want to add some test cases to validate that the custom series works: assert sum_series(5, 5, 7) == 50

ervbrian: Added a few more assertions.

students/BrianE/session02/GridPrinter.py
ervbrian: I've added an example of how my function supports a single parameter (box_size). The default value of 2 for grid_size supports calling grid(5). 

Let me know if you want a completely separate function with only a single parameter defined and I'll add that, instead.

ervbrian: Went ahead and added the initial print_grid() function just in case.

nairraghav:
Nice job Brian, this looks great!

PR #56, opened 2019-04-20
Lesson 02 Assignment

miotke:
Here are the assignments for lesson 02.
I also took notes for the reading and videos which can be ignored, sorry for including them in the PR.

I did run into a few issues that I need some guidance on since I wasn't able to attend class on 04/16 due to being out of town :( 

### series.py
* I struggled to get the Fibonacci series and Lucas series to successfully run in the `sum_series()` function. I wrote it before realizing the `series_template.py` existed. It looks like we are down similar paths with it. When I try and change the parameter values it get a failure. 

### grid_printer.py
* I can pass in arguments like `create_grid(3, 3)` and that will create a perfect 3x3 grid. However if I pass in `create_grid(2,3)` I do not get a 2x3 grid as expected. I'm know there's a simple solution to this problem I just couldn't seem to figure it out. :(


students/AndrewMiotke/session02/grid_printer.py
```diff
@@ -0,0 +1,51 @@
...
+def create_grid(beams, posts):
+    """
+    Returns a grid based on the parameters of this function
+    :param arg 2: define the number of beams(x axis) required to complete a cell
+    :param arg 3: define the number of posts(y axis) required to complete a cell
+
+    Example: create_grid(10, 10) would create a 10x10 grid
+    """
+    for _ in range(posts):
+        create_beam(beams)
```
nairraghav: This is a good start but do note that you're hardcoding the size of each cell. 

The initial print_grid should always create a grid of 4 cells with the total height/length being the parameter we give it. For example: print_grid(9) should return a 4x4 grid where the entire length/height (2 cells worth) is 9. Hint: The middle '+' will handle it such that the two cells are of 4 length each. 

The final print_grid should take two parameters.  The first parameter determines the nxn grid size and the second parameter determines the size of each cell. So for example, print_grid(5,3) should return a 5x5 grid where each cell is 3x3 (3 '-' across and 3 '|' long )

students/AndrewMiotke/session02/series.py
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env python3
```
nairraghav, line 1: This looks great!

nairraghav:
Awesome job so far, your fizzbuzz and series are complete. Please take a look at my notes for the grid printer.

students/AndrewMiotke/session02/series.py
miotke: Awesome, thank you! I didn't think it fulfilled all of requirements @nairraghav

miotke:
@nairraghav thank you for the feedback. I think I fixed the grid_printer file. I still had trouble getting the x axis to be dynamic to the function argument. :(

students/AndrewMiotke/session02/grid_printer.py
miotke: Thank you for the feedback. I ended up writing a new file and I believe it's working now. Should be in the latest commit.

students/AndrewMiotke/session02/series.py
```diff
@@ -0,0 +1,63 @@
...
+    if n == 0:
+        return 2
+    elif n == 1:
+        return 1
+    else:
+        return lucas(n-2) + lucas(n-1)
+
+print(lucas(4))
+
+def sum_series(n, t=0, r=1):
```
PythonCHB, line 26: perfect!

students/AndrewMiotke/session02/grid_printer.py
```diff
@@ -0,0 +1,64 @@
...
+create_grid(5)
+
+def print_grid(cells, size):
+    """
+    Returns a Xx2 grid with flexible cell height and width
+    :param arg 0: defines the number of cells along the y axis
+    :param arg 1: defines the height('|') and width('-') of each cell
+   """
+    for _ in range(cells):
+        create_beam(size)
```
nairraghav, line 58: You're pretty close here! One way to think about it is to modify your create_beam function so that it takes another variable to determine how many "mini-beams" make up the whole beam. 

The way I think about it is to minimize the function to print just 1 beam (+----+). Then using a for loop and your new variable, you repeat it so you can print multiple beams. 

Ex: 
- create_beam(2, 1) returns +--+
- create_beam(2,2) returns +--+--+

Once that's updated, you can call that here and it should work! Just remember that you will need to update it in your create_grid() as well. But in that one, since we always have a 2x2, you can use create_beam(size, 2)

students/AndrewMiotke/session02/series.py
nairraghav, line 26: Nice!

students/AndrewMiotke/session02/series.py
```diff
@@ -0,0 +1,63 @@
...
+    # test lucas
+    assert lucas(0) == 2
+    assert lucas(1) == 1
+    assert lucas(4) == 7
+
+    # test if sum_series is equal to fibonacci
+    assert sum_series(5) == fibonacci(5)
+
+    # test if sum_series is equal to lucas
+    # assert sum_series(5) == lucas(5)
```
nairraghav, line 61: you will want to update your initial parameters to test against lucas:
assert sum_series(5, 2, 1) == lucas(5)

PR #57, opened 2019-04-20
series ex. complete

jchando1:


students/JasneetChandok/Lesson02/series.py
```diff
@@ -0,0 +1,78 @@
...
+    :param n0=0: value of zeroth element in the series
+    :param n1=1: value of first element in the series
+    
+    This function should generalize the fibonacci() and the lucas(),
+    so that this function works for any first two numbers for a sum series.
+    Once generalized that way, sum_series(n, 0, 1) should be equivalent to fibonacci(n).
+    And sum_series(n, 2, 1) should be equivalent to lucas(n).
+    """
+    if n0 == False:
+        if n > 1:
```
nairraghav: Remember that if n0 == 0 (False), you can just use your Fibonacci function. Otherwise, you are essentially re-doing the work.

students/JasneetChandok/Lesson02/series.py
```diff
@@ -0,0 +1,78 @@
...
+            return fibonacci(n - 1) + fibonacci(n - 2)
+        if n == 1:
+            return n1
+        elif n <= 0:
+            return n0
+    else:
+        if n > 1:
+            return lucas(n - 1) + lucas(n - 2)
```
nairraghav: When you call lucas, you're always setting your n0 to 2 and n1 to 1. You will never reach the elif n == 1 or n <= 0 since the lucas functions are recursive. 

Hint: You may want to generalize the lucas function (or implement it here itself) where n0 and n1 can be anything. You can use this generalized new logic to handle any n0 and n1 inputs

nairraghav:
The fibonacci and lucas functions look great. Please refer to the comments for the generalized sum_series which should handle any n0 and n1.

nairraghav:
Nice job, your generalized logic works! As a side note, this logic also handles the n0,n1 = 0,1 and n0,n1 = 2,1. So we technically don't even need the if statements to call the fib/lucas functions. By generalizing it, it will handle all initial values

students/JasneetChandok/Lesson02/series.py
```diff
@@ -0,0 +1,89 @@
...
+    Once generalized that way, sum_series(n, 0, 1) should be equivalent to fibonacci(n).
+    And sum_series(n, 2, 1) should be equivalent to lucas(n).
+
+    sum_series(n, 3, 2) should generate another series with no specific name
+
+    The defaults are set to 0, 1, so if you don't pass in any values, you'll
+    get the fibonacci sercies
+    """
+    if n0 == 0 and n1 == 1:
+        return fibonacci(n)
```
PythonCHB, line 46: the idea here was to do the opposite -- this code can compute the fibonacci and lucas numbers, so no need to call them here. But you could refactor them to call this function -- then there would only be the logic in one place.

PR #58, opened 2019-04-21
adding more test variants to enumerate the implicit and explicit call…

JoeNunnelley:
… variants for fibonacci and to provide better explanatory comments

### DIFF
--- a/students/JoeNunnelley/session02/series.py
+++ b/students/JoeNunnelley/session02/series.py
@@ -32,6 +32,8 @@ def lucas(n):
 
 def sum_series(n, first=0, second=1):
     """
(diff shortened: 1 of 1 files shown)


nairraghav:
Hey Joe, it looks like this is still missing the generalized function. Your if statement looks to either do fibonacci or lucas series. Imagine a series that doesn't fit either.

Ex: sum_series(5, 2, 4)
# 2 4 6 10 16 26
The 5th in the series would be 26 (since 0th is 2). In your exising code, since 2 != 0, it goes to the lucas function which hardcodes the n0 = 2, and n1 = 1.

nairraghav:
Nice job Joe! This generalized version looks good

PR #59, opened 2019-04-21
adding grid printer

pberhe:
only grid printer assignment

### DIFF
--- a/students/PeterB/Session 2/Grid.py
+++ b/students/PeterB/Session 2/Grid.py
@@ -1,0 +1,23 @@
+plus = '+'
+minus = '- '
+bar = '|'
+space = ' '
+
+# print_grid2(3,4)Width = 4
+
+def print_row(length):
+    print(plus + space + minus * length + plus+space + minus * length + plus)
+
+def print_column(length):
+    print(bar + space+space*length *2 + bar + space+space*length*2 + bar)
+
+
+def grid_print (length):
+    for _ in range(2):
+        print_row(length)
+        for _ in range(length):
+            print_column(length)
+    print_row(length)
+
+grid_print(3)
+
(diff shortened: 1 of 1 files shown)


students/PeterB/Session 2/Grid.py
```diff
@@ -0,0 +1,32 @@
+plus = '+'
```
PythonCHB, line 1: you really don't want a space in filenames -- it makes using the command line challenging. you can change it in git like so:

git mv "Session 2" Session2

you need to quotes so that git doesn't interpret the space as new name.

students/PeterB/Session 2/Grid.py
```diff
@@ -0,0 +1,32 @@
...
+minus = '- '
+bar = '|'
+space = ' '
+
```
PythonCHB, line 5: I like the idea of naming the characters, so you could change them in one place.

But better tpo use more generic names, like "corner" in stead of "plus".

This way, it would get kind of confusing to do:

plus = "*"

and if you aren't going to change them, why give them a name?

students/PeterB/Session 2/Grid.py
```diff
@@ -0,0 +1,32 @@
...
+# print_grid2(3,4)Width = 4
+
+def print_row(length):
+    print(plus + space + minus * length + plus+space + minus * length + plus+space + minus * length+ plus)
```
PythonCHB, line 9: I can't help thinking there could be a more compact way to do this...

PythonCHB: But the bigger problem is that this will always makeit 3 cells wide -- we want that to adjustable, so you can make a 4x4 grid, for instance.

PythonCHB (changes requested):
close enough -- I'll merge and give you credit, but if you have a change to make the number f columns adjustable, that would be better.

PR #61, opened 2019-04-22
lesson2 assignment

momosally:
Please check the three exercises!

### DIFF
--- a/students/Sally_Shen/lesson2/fizzBuzz.py
+++ b/students/Sally_Shen/lesson2/fizzBuzz.py
@@ -1,0 +1,23 @@
+''''Fizz Buzz
+Write a program that prints the numbers from 1 to 100 inclusive.
+But for multiples of three print “Fizz” instead of the number.
+For the multiples of five print “Buzz” instead of the number.
+For numbers which are multiples of both three and five print “FizzBuzz” instead.
+'''
+
+def fizzBuzz(n):
+
+    output = []
+    for i in range(1,n+1):
+        if i % 3 == 0 and i % 5 == 0:
+            output.append("FizzBuzz")
+        elif i % 3 == 0:
+            output.append('Fizz')
+        elif i % 5 == 0:
+            output.append('Buzz')
+
+        else:
+            output.append(str(i))
+    return output
+
+print(fizzBuzz(20))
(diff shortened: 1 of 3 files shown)


students/Sally_Shen/lesson2/fizzBuzz.py
```diff
@@ -0,0 +1,23 @@
...
+        if i % 3 == 0 and i % 5 == 0:
+            output.append("FizzBuzz")
+        elif i % 3 == 0:
+            output.append('Fizz')
+        elif i % 5 == 0:
+            output.append('Buzz')
+
+        else:
+            output.append(str(i))
+    return output
```
PythonCHB, line 21: I like how you didn't do the printing in the function -- nice!

but maybe a way to print the list a little nicer?

```
for line in fizzBuzz(20):
    print line
```

But nice work!

students/Sally_Shen/lesson2/gridPrinter.py
```diff
@@ -0,0 +1,39 @@
...
+    edge += '\n'
+    row += '\n'
+
+    result = edge
+    for i in range(rowCols):
+        for j in range(dashes):
+            result += row
+        result += edge
+    print(result)
+
```
PythonCHB, line 22: they are a little tall and skinny -- the example has spaces between the dashes.

But you get the idea.

-CHB

students/Sally_Shen/lesson2/series .py
```diff
@@ -0,0 +1,30 @@
+'''''Goal: The Fibonacci Series is a numeric series starting with the integers 0 and 1.
```
PythonCHB, line 1: careful! looks like you have a space before the dot in the filename -- that could cause problems iwth the command line.

you can rename it in git with:

git mv "series .py" series.py

you need the quotes so it will read the space as part of the filename.

students/Sally_Shen/lesson2/series .py
```diff
@@ -0,0 +1,30 @@
...
+
+def lucas(n):
+
+    a,b = 2,1
+    for _ in range (n):
+        a,b = b, a+b
+    return a
+
+print(lucas(0))
+
```
PythonCHB, line 26: looks like you aren't done here -- will lucas return the number for n=0 and n=1

add some more asserts to test the results!

And you still need the sum-series function...

See the template linked in the exercise description.

PythonCHB (approved):
print grid and fizz buzz look good -- full credit for those.

You still need to finish sum_series.

PR #64, opened 2019-04-23
New pull request for Lesson 02 files

BBrumleve:
Files are:
 1) grid_printer.py
 2) fizz_buzz.py
 3) series.py


students/BrianB/lesson01/sum_double.py
```diff
@@ -0,0 +1,19 @@
...
+#
+# sum_double(1, 2) → 3
+# sum_double(3, 2) → 5
+# sum_double(2, 2) → 8
+#------------------------------------------
+
+def sum_double(a,b):
+    if a != b:
+        return(a+b)
+    elif a==b:
```
PythonCHB, line 14: if a!= b failed, then a == b will always pass (at least for numbers). so don't need the elif.

students/BrianB/lesson02/fizz_buzz.py
```diff
@@ -0,0 +1,40 @@
...
+#   printed and for multiples of 3 adn 5,
+#   "FizzBuzz" is printed.
+#--------------------------------------------------
+
+# Global variables set the
+# range to be looped through
+n = 100
+i = 0
+
+while i < n:
```
PythonCHB, line 19: better to use a for lopo with range():

```
for i in range(n):
    ...
```

students/BrianB/lesson02/fizz_buzz.py
```diff
@@ -0,0 +1,40 @@
...
+        print("FizzBuzz")
+    elif i % 3 == 0:
+        print("Fizz")
+    elif i % 5 == 0:
+        print("Buzz")
+    else:
+        # if a value isn't divisible by 3 or 5
+        # then its value ends up being printed
+        # by this print statement
+        print(i)
```
PythonCHB, line 40: good, solid solution.

-CHB

students/BrianB/lesson02/grid_printer.py
```diff
@@ -0,0 +1,162 @@
...
+    # establishing the range is an
+    # integer and ensuring integration
+    # with the col() function to
+    # ultimately print the correct
+    # width rows: plus signs and dashes
+    # and the correct width columns:
+    # vertical lines and spaces.
+    if b % 2 == 0:
+        x = int(b) / 2
+    elif b % 2 == 1:
```
PythonCHB, line 37: same as before, no need for a elif if it's always going to be True.

you could also use // for "floor division", rather than checking for even or odd:

x = b // 2

```
In [17]: for i in range(8): 
    ...:     print(i // 2) 
    ...:                                                                        
0
0
1
1
2
2
3
3
```

students/BrianB/lesson02/series.py
```diff
@@ -0,0 +1,117 @@
...
+# Fibonacci Series
+#   This function assigns optional parameters (OP)
+#   specific to the Fibonacci Series.  These
+#   optional parameters represent the first two
+#   values of F(0) and F(1).  The math for
+#   developing the Fibonacci Series is performed
+#   in the sum_series() function, which is called
+#   to by the fibonacci() function.
+def fibonacci(n, a=0, b=1):
+    return sum_series(n, a, b)
```
PythonCHB, line 24: yes!

except the Fibonacci series always starts with (0, 1), so this should not take optional a and b parameters.

This is essentially exactly the same as the series function ...

students/BrianB/lesson02/series.py
```diff
@@ -0,0 +1,117 @@
...
+#    & Other Series
+#   This function assigns optional parameters (OP)
+#   specific to the Fibonacci Series by default.
+#   If the optional parameters are overwritten by
+#   with new inputs, the sum_series will create an
+#   entirely new series unless the input parameters
+#   happen to match those of the Lucas Numbers.  Then
+#   the Lucas Numbers will be created.
+def sum_series(n, a=0, b=1):
+    a, b = a, b
```
PythonCHB, line 49: a and b already equal a and b .... no need for this line.

students/BrianB/lesson02/series.py
```diff
@@ -0,0 +1,117 @@
...
+        # A Lucas Number
+        assert sum_series(2, 2, 1) == 3
+        assert sum_series(4, 2, 1) == 7
+        # A series all its own
+        assert sum_series(8, 3, 3) == 102
+        assert sum_series(9, 3, 3) == 169
+        assert sum_series(10, 3, 3) == 267
+        assert sum_series(11, 3, 3) == 432
+        assert sum_series(12, 3, 3) == 699
+        assert sum_series(13, 3, 3) == 1131
```
PythonCHB, line 113: did you compute this by hand?  ;-)

PythonCHB (approved):
Nice work!

PR #65, opened 2019-04-23
'session2'

pberhe:
grid, series, & fizzbuzz

### DIFF
--- a/students/PeterB/Session2/fizzbuzz.py
+++ b/students/PeterB/Session2/fizzbuzz.py
@@ -1,0 +1,11 @@
+for fizzbuzz in range(100):
+    if fizzbuzz % 3 == 0 and fizzbuzz % 5 == 0:
+        print("fizzbuzz")
+        continue
+    elif fizzbuzz % 3 == 0:
+        print("fizz")
+        continue
+    elif fizzbuzz % 5 == 0:
+        print("buzz")
+        continue
+    print(fizzbuzz)
--- a/students/PeterB/Session2/series2.py
+++ b/students/PeterB/Session2/series2.py
@@ -1,0 +1,31 @@
+# fibonacci
+
+a, b = 0,1
+while b < 56:
+    print(b)
+    a, b = b, a+b
+
+# lucas
+
+a, b = 2,1
+while b < 56:
+    print(b)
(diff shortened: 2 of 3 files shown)


students/PeterB/Session2/fizzbuzz.py
```diff
@@ -0,0 +1,11 @@
+for fizzbuzz in range(100):
+    if fizzbuzz % 3 == 0 and fizzbuzz % 5 == 0:
+        print("fizzbuzz")
+        continue
```
nairraghav: Instead of using continues and elifs, you can achieve the same with:

if fizzbuzz % 3 == 0 and fizzbuzz % 5 == 0:
        print("fizzbuzz")
if fizzbuzz % 3 == 0:
        print("fizz")
if fizzbuzz % 5 == 0:
        print("buzz")

students/PeterB/Session2/series2.py
```diff
@@ -0,0 +1,31 @@
...
+
+# sum_series
+# one required user , two optional parameters
+
+# n, b, c = ,0,1
+# while n < 56:
+#     print(n)
+
+
+def sum_series(n, b=0, c=1):
```
nairraghav, line 23: sum_series looks good!

students/PeterB/Session2/series2.py
```diff
@@ -0,0 +1,31 @@
+# fibonacci
+
+a, b = 0,1
```
nairraghav, line 3: For the Fibonacci and Lucas series, you would want to create a function similar to how sum_series does it.

students/PeterB/Session2/fizzbuzz.py
```diff
@@ -0,0 +1,11 @@
...
+    elif fizzbuzz % 3 == 0:
+        print("fizz")
+        continue
+    elif fizzbuzz % 5 == 0:
+        print("buzz")
+        continue
+    print(fizzbuzz)
```
nairraghav, line 11: We only want to print out the number if it doesn't fit the categories above. In this case, it is always printing.

Note: You don't want the continues in the if statements. Once we hit one of the "ifs", we don't want to do anything else for that number

PR #67, opened 2019-04-23
session2.1

pberhe:
fizzbuzz/series updated

### DIFF
--- a/students/PeterB/Session2/fizzbuzz.py
+++ b/students/PeterB/Session2/fizzbuzz.py
@@ -1,11 +1,8 @@
 for fizzbuzz in range(100):
     if fizzbuzz % 3 == 0 and fizzbuzz % 5 == 0:
         print("fizzbuzz")
-        continue
-    elif fizzbuzz % 3 == 0:
-        print("fizz")
-        continue
-    elif fizzbuzz % 5 == 0:
-        print("buzz")
-        continue
+    fizzbuzz % 3 == 0
+    print("fizz")
+    fizzbuzz % 5 == 0
+    print("buzz")
     print(fizzbuzz)
--- a/students/PeterB/Session2/series.py
+++ b/students/PeterB/Session2/series.py
@@ -1,0 +1,101 @@
+# fibonacci number = 0,1,1,3,5,8,13, 21
+
+def fibonacci(n):
+    a,b = 0, 1
+    for i in range(n-1):
+        a, b = b, a+b
+    return a
+
+for i in range(0,10):
+    print ((fibonacci(n))
+
+
+def lucas(n):
+    a, b = 2, 1
+    for i in range(n - 1):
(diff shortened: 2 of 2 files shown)


students/PeterB/Session2/fizzbuzz.py
```diff
@@ -1,11 +1,8 @@
...
     if fizzbuzz % 3 == 0 and fizzbuzz % 5 == 0:
         print("fizzbuzz")
-        continue
-    elif fizzbuzz % 3 == 0:
-        print("fizz")
-        continue
-    elif fizzbuzz % 5 == 0:
-        print("buzz")
-        continue
+    fizzbuzz % 3 == 0
```
PythonCHB, line 4: make sure to test the code! this prints a whole bunch too may fizzes and buzzes....

you didn't need the continue, but you DO need the `elif`s

students/PeterB/Session2/series.py
```diff
@@ -0,0 +1,101 @@
+# fibonacci number = 0,1,1,3,5,8,13, 21
+
+def fibonacci(n):
+    a,b = 0, 1
+    for i in range(n-1):
+        a, b = b, a+b
+    return a
```
PythonCHB, line 7: almost right -- it's supposed to start with 0, 1, 1, 2, ....

butyou get an extra 0:

```
In [2]: def fibonacci(n): 
   ...:     a,b = 0, 1 
   ...:     for i in range(n-1): 
   ...:         a, b = b, a+b 
   ...:     return a 
   ...:                                                                         

In [3]: fibonacci(0)                                                            
Out[3]: 0

In [4]: fibonacci(1)                                                            
Out[4]: 0

In [5]: fibonacci(2)                                                            
Out[5]: 1
```

PythonCHB: and by returning a, you are computing a final b that you don't use. not a big deal here, but in generaly, you don't want to compute anything extra.

students/PeterB/Session2/series.py
```diff
@@ -0,0 +1,101 @@
...
+        a, b = b, a + b
+    return a
+
+    for i in range(0, 10):
+        print((lucas(n))
+
+#
+# # sum_series
+#
+#def sum_series(n, b=0, c=1):
```
PythonCHB, line 25: this is commented out ?!?

students/PeterB/Session2/series.py
```diff
@@ -0,0 +1,101 @@
...
+    if n < 0:
+        return None
+    elif n == 0:
+        return b
+    elif n == 1:
+        return c
+    else:
+        return sum_series(n - 1, b, c) + sum_series(n - 2, b, c)
```
PythonCHB, line 33: but this function looks good -- and with recursion!

PythonCHB (approved):
calling it good -- nice work.

PR #68, opened 2019-04-23
Added grid.py, FizzBuzz.py, series_fibonacci_lucas.py

RimleeTalukdar44:
1. deleted grid1.py
2. Added grid.py 
3. Added FizzBuzz.py
4. Added series_fibonacci_lucas.py


students/RimleeT/lesson02/grid.py
```diff
@@ -0,0 +1,68 @@
...
+Part 2 : Write a function print_grid(n) that takes one integer argument and prints a grid just like before, BUT the size of the grid is given by the argument.
+"""
+def print_grid(n):
+	size_unit = (n-1)//2
+	plus = '+'
+	minus = '-'*size_unit
+	line = '|'
+	space = ' '*size_unit
+	print ((plus + minus)*2 + plus)
+	for i in range(0,size_unit):
```
nairraghav, line 33: Overall, your grindprint looks great! Just a note, when using range and your initial value is 0, you do not need to specify it. range (0, 3) is equivalent to range(3)

students/RimleeT/lesson02/series_fibonacci_lucas.py
```diff
@@ -0,0 +1,54 @@
...
+It should have one required parameter and two optional parameters. The required parameter will determine which element in the series to print. The two optional parameters will have default values of 0 and 1 and will determine the first two values for the series to be produced.
+Calling this function with no optional parameters will produce numbers from the fibonacci series (because 0 and 1 are the defaults).
+Calling it with the optional arguments 2 and 1 will produce values from the lucas numbers.
+"""
+def sum_series(n, n0=0, n1=1):
+	if n0==2 and n1==1:
+		print("Calls lucas")
+		x=lucas(n)
+		return x
+	else:
```
nairraghav, line 28: We don't want this to be always because our choices aren't always between lucas and fibonacci. You should be using generalized logic that can handle any input:

ex:
sum_series(3, 1, 1) = 3 (this needs to be implemented, it is very similar to the logic for fib/lucas)
    1, 1, 2, 3,
sum_series(3, 0, 1) = fib(3)
sum_series(3, 2, 1) = lucas(3)

nairraghav:
It looks like you have a merge conflict. Pull down the upstream code, fix the merge issues and then commit/push your changes. 

Let me know if you have any trouble with this, I can help out as well. 

Otherwise, everything looks pretty good, please address my comment for the series.

PR #69, opened 2019-04-23
revised series

momosally:
All passed! Please check the series problem. Add the sum_series and more tests.


nairraghav:
Nice job Sally!

PR #71, opened 2019-04-25
Adding Session03 files

ervbrian:
Added SlicingLab.py

### DIFF
--- a/students/BrianE/session03/SlicingLab.py
+++ b/students/BrianE/session03/SlicingLab.py
@@ -1,0 +1,101 @@
+#!/usr/env python3
+
+"""
+SlicingLab.py
+
+Sequence manipulation to achieve the following:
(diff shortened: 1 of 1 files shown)


students/BrianE/session03/SlicingLab.py
```diff
@@ -0,0 +1,101 @@
...
+
+def remove_first_last_4_every_other(sequence):
+    """
+    Return sequence with first/last four removed along with every other element removed.
+
+    :param sequence: input sequence (string, list, tuple, etc)
+    :return: sequence with first/last four removed along with every other element removed
+    """
+
+    return sequence[4:-4][::2]
```
PythonCHB, line 45: you can put the step right in with the first slice: `[4:-4:2]`

PythonCHB (approved):
this all looks good -- nice job!

PR #73, opened 2019-04-26
lesson 3: sstrformat_lab and slicing_lab questions...

BBrumleve:
Chris or Ron, 
On StrFormat_Lab.py, Task Three I'd like to know if my "Task Three_2" is what you're looking for.  If not I've been working to try and create some sort of function that can accept a changing number of parameters, as what I see in the exercise notes ...a function accepting...formatter(3,2,5) and formatter(3,2,5,8,9).  Any advice on what i currently have? I'm stuck on this.

On Slicing_Lab,py, no questions yet.  It just got pulled into this pull request with StrFormat_Lab.py

Cheers,
BrianB

### DIFF
--- a/students/BrianB/lesson03/slicing_lab.py
+++ b/students/BrianB/lesson03/slicing_lab.py
@@ -1,60 +1,24 @@
-print("hello world")
-
-#---------------------------------------------------------------------
-# Goal:
-# Get the basics of sequence slicing down.
-#
-# Tasks:
-#  Write some functions that take a sequence as an argument, and
-#  return a copy of that sequence:
-#    1* with the first and last items exchanged.
-#           SOLUTION == a_string[-1:] + a_string[1:-1] + a_string[:1] &
-#                       a_tuple[-1:] + a_tuple[1:-1] + a_tuple[:1]
-#    2* with every other item removed.
-#           SOLUTION == a_string[0:-1:2] &
-#                       a_tuple[0:-1:2]
-#    3* with the first 4 and the last 4 items removed, and then
-#       every other item in the remaining sequence.
-#           SOLUTION == a_string[4:-4:2] &
-#                       a_tuple[4:-4:2]
-#    4* with the elements reversed (just with slicing).
-#           SOLUTION == a_string[::-1] &
-#                       a_tuple[::-1]
-#    5* with the last third, then first third, then the middle
-#       third in the new order.
-#           SOLUTION == a_string[11:16] + a_string[0:5] + a_string[5:11] &
-#                       a_tuple[4:6] + a_tuple[0:2] + a_tuple[2:4]
-#
-# NOTE: These should work with ANY sequence – but you can use
-#       strings to test, if you like.
-#
-#
-# Tests:
-# a_string = "this is a string"
-# a_tuple = (2, 54, 13, 12, 5, 32)
-#
-# assert exchange_first_last(a_string) == "ghis is a strint"
-# assert exchange_first_last(a_tuple) == (32, 54, 13, 12, 5, 2)
-#---------------------------------------------------------------------
+#!/usr/bin/env python3
 
 a_string = "this is a string"
 a_tuple = (2, 54, 13, 12, 5, 32)
 
 
 def exchange_first_last(seq):
-    """Returns an exchange of the first and last elements."""
+    """Returns an exchange of the first element with the last element."""
     a_new_sequence = seq[-1:] + seq[1:-1] + seq[:1]
     return a_new_sequence
 
 
 def every_other(seq):
-    """Returns every other element."""
+    """Returns every other element in the sequence."""
     a_new_sequence = seq[0:-1:2]
     return a_new_sequence
 
 
 def first_last_four_every_other(seq):
-    """Returns the first four, the last four and every other element."""
+    """Returns the first four elements, then the last four\
+    elements and every other element."""
     a_new_sequence = seq[4:-4:2]
     return a_new_sequence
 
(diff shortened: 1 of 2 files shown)


PythonCHB:
On a phone on a bus — so real quick:

Be sure to watch the video on Canvas on dynamic string formatting.

Your function doesn’t need to accept a variable number of arguments (it could, but we’ll cover that in a couple weeks) Rather, it should accept a triple with an arbitrary number of elements:

Formatter((1,6,3,8,2))

Or

numbers = (7,3,9,1,4,6)

Formatter(numbers)

Hope that helps.

students/BrianB/lesson03/strformat_lab.py
```diff
@@ -0,0 +1,88 @@
...
+#     tup =  "the 3 numbers are: {:d}, {:d}, {:d}".format(x, y, z)
+#     return tup
+#
+#
+# print((formatter(3, 2, 5)))
+
+
+# Task Three_3
+def formatter(tup):
+    tup = str(tup)
```
PythonCHB, line 79: you don't want to make a string out of it!

(well, maybe, that would be one way to kludge this requirement, but that's not the goal here :-)

students/BrianB/lesson03/strformat_lab.py
```diff
@@ -0,0 +1,88 @@
...
+    counter = 0
+    for i in tup:
+        counter += 1
+    tup = "the " + str(counter) + " numbers are: {:d}, {:d}, {:d}".format(*str)
+    return tup, counter
+print(formatter(2, 3, 5))
```
PythonCHB, line 85: so this gives an error, 'cause formatter is expecting one argument -- a tuple of numbers. But here you are passing in three separate numbers.

I suspect his is a work in progress -- please make a note when you have an open PR -- either on the PR, or in the commit message.

The code actually doesn't run for other reasons as well ...

students/BrianB/lesson03/strformat_lab.py
```diff
@@ -0,0 +1,88 @@
...
+# def formatter(x, y, z):
+#     tup =  "the 3 numbers are: {:d}, {:d}, {:d}".format(x, y, z)
+#     return tup
+#
+#
+# print((formatter(3, 2, 5)))
+
+
+# Task Three_3
+def formatter(tup):
```
PythonCHB, line 78: This is clearly a work in progress, but I can see that you are headed in the right direction.

but lots o' notes :-)

students/BrianB/lesson03/strformat_lab.py
```diff
@@ -0,0 +1,88 @@
...
+    tup = str(tup)
+    counter = 0
+    for i in tup:
```
PythonCHB, line 81: if you want to loop through a sequence, and keep track of the index, you can use "enumerate":

```
for counter, item in enumerate(tup):
    ...
```

PythonCHB, line 81: now that tup is a string, looping through it will give you each character, which you really don't want :-)

students/BrianB/lesson03/strformat_lab.py
```diff
@@ -0,0 +1,88 @@
...
+        counter += 1
+    tup = "the " + str(counter) + " numbers are: {:d}, {:d}, {:d}".format(*str)
```
PythonCHB, line 83: if you are using format, then the "the" and " numbers are" could/should be all in the format string.

you really don't want `*str` -- `str` is the string object :-)

once you've formatted it, you probably shouldn't call it "tup" anymore.

PythonCHB (changes requested):
Big note about process:

Run your code as often as you can as you work on it -- you should rarely have more than one line of code that won't even run. And it shouldn't stay broken for long:

a - have an assert, or print, or something that runs your code (you've got that)
b - add / change as little as possible (usually one line)
c - run it, and see if the addition / change is doing what you want
d - back to (b)

As you get more experienced, you'll start doing more than one line at a time, but it should always be pretty small chunks.

And you almost never should commit code that doesn't run.

students/BrianB/lesson03/slicing_lab.py
```diff
@@ -0,0 +1,56 @@
...
+
+def exchange_first_last(seq):
+    """Returns an exchange of the first element with the last element."""
+    a_new_sequence = seq[-1:] + seq[1:-1] + seq[:1]
+    return a_new_sequence
+
+
+def every_other(seq):
+    """Returns every other element in the sequence."""
+    a_new_sequence = seq[0:-1:2]
```
PythonCHB, line 15: careful -- putting the -1 in there means you won't get the last one -- if there are an odd number of items, yo'll miss one.

and you don't need the 0 -- that's implied. So this can simply be: `[::2]`

students/BrianB/lesson03/slicing_lab.py
```diff
@@ -0,0 +1,56 @@
...
+    return a_new_sequence
+
+
+def first_last_four_every_other(seq):
+    """Returns the first four elements, then the last four\
```
PythonCHB, line 20: your description is a bit off, but the solution is correct :-)

students/BrianB/lesson03/slicing_lab.py
```diff
@@ -0,0 +1,56 @@
...
+    elements and every other element."""
+    a_new_sequence = seq[4:-4:2]
+    return a_new_sequence
+
+
+def reverse_order(seq):
+    """Returns the elements in reverse order"""
+    a_new_sequence = seq[::-1]
```
PythonCHB, line 28: bingo!

students/BrianB/lesson03/slicing_lab.py
```diff
@@ -0,0 +1,56 @@
...
+    return a_new_sequence
+
+
+def new_order(seq):
+    """Returns a New Order: the last three elements, then\
+    the first three elements and then the middle three elements"""
+    a_length = len(seq) // 3
+    a_new_sequence = seq[a_length*2:] + seq[:a_length] + seq[a_length:a_length*2]
```
PythonCHB, line 36: yup.

PythonCHB (approved):
slicing looks great!

PythonCHB:
merging and giving credit for slicing lab.

Do a new PR when you're ready for the formatting lab.

PR #75, opened 2019-04-27
listslab update

trayffs:


students/TylerRay/session3/listslab.py
```diff
@@ -0,0 +1,69 @@
...
+    4) Add fruit to the beginning of the list
+
+    5) Delete the first fruit
+
+    6) Delete fruit by number
+
+    7) Exit Program
+    """)
+    strChoice = str(input("Which option would you like to perform? [1 to 4] - "))
+    print()
```
PythonCHB, line 22: if you call strChoice.strip() here, then you only have to call it once.

students/TylerRay/session3/listslab.py
```diff
@@ -0,0 +1,69 @@
...
+
+    4) Add fruit to the beginning of the list
+
+    5) Delete the first fruit
+
+    6) Delete fruit by number
+
+    7) Exit Program
+    """)
+    strChoice = str(input("Which option would you like to perform? [1 to 4] - "))
```
PythonCHB, line 21: You are using "Hungarian Notation: here. This is generally not recommend in Python. You can read ahead a bit about naming things here:

https://uwpce-pythoncert.github.io/PythonCertDevel/modules/NamingThings.html

PythonCHB (approved):
Nice work! you've got the list manipulations down.

PR #77, opened 2019-04-28
Adding list_lab, strformat and MailRoom.py files

ervbrian:


students/BrianE/session03/list_lab.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+def cleanup_list(fruits):
+    """
+    Remove fruits from list if user does not like
+    :param fruits: list of fruits
+    :return: None
+    """
+
+    for fruit in fruits:
```
nairraghav: You will want to be careful here. Everytime we say no and remove a fruit, we jump up an extra one (because we are not accounting for the fruit we just removed). Try this out and see how it acts. 

Hint: Try to iterate through the fruits list where you can control the iterator, otherwise it will jump an extra spot when you don't want it to.

ervbrian: Corrected, but I don't think this is the best solution. I attempted to subtract 1 from the value of i when the user answers "no". The next item in the list was still skipped and I cannot figure out why.

Instead, I've created a new list of items to delete and I iterate through them afterwards, removing from the original list. Let me know what you think.

students/BrianE/session03/list_lab.py
```diff
@@ -0,0 +1,138 @@
...
+            if answer == 'yes':
+                break
+            elif answer == 'no':
+                deleted_fruits.append(fruits[i])
+                break
+            else:
+                print(f"Please choose from the following: yes|no")
+                continue
+
+    for fruit in deleted_fruits:
```
nairraghav, line 55: This does work! You will just want to keep in mind that when working with large datasets, creating a secondary list could consume a lot of memory.

students/BrianE/session03/list_lab.py
nairraghav: My apologies, you would not be able to modify the iterator value itself, you would likely need to keep another variable to track it. Your way of solving this works as well, just keep the note about memory in mind!

nairraghav:
Nice job overall, this looks great Brian!

PR #78, opened 2019-04-28
Completed slicing lab exercise

jchando1:


students/JasneetChandok/Lesson03/slicinglab.py
```diff
@@ -0,0 +1,72 @@
...
+	return new_sequence
+
+
+def elements_reversed(seq):
+	'''elements reversed (just with slicing)'''
+	new_sequence = seq[::-1]
+	return new_sequence
+
+
+def first_last_middle_thrid(seq):
```
nairraghav, line 30: This is very close except you are taking the last third and reversing it. We want last third - first third - middle third. You will want to update this to fit as expected. Your test case will also need to be updated

students/JasneetChandok/Lesson03/slicinglab.py
```diff
@@ -0,0 +1,72 @@
...
+	assert first4_last4_eveyother(a_tuple) == ()
+	assert first4_last4_eveyother(a_list) == []
+
+	#Test - elements reversed (just with slicing)
+	assert elements_reversed(a_string) == "gnirts a si siht"
+	assert elements_reversed(a_tuple) ==(32, 5, 12, 13, 54, 2)
+	assert elements_reversed(a_list) ==['dog', 5, 12, 13, 'age', 2]
+
+	#Test - last third, then first third, then the middle third in the new order
+	assert first_last_middle_thrid(a_string) == "gnithiis a "
```
nairraghav, line 65: Your test case should return something similar but the first third should not be reversed

nairraghav:
I am merging this for now since you have everything else solved, please submit another PR with the fix for your thirds slice

PR #79, opened 2019-04-28
Both list lab and slicing lab assignment have been completed

jchando1:


students/JasneetChandok/Lesson03/listlab.py
```diff
@@ -0,0 +1,100 @@
...
+Ask the user for input displaying a line like “Do you like apples?” for each fruit in the list (making the fruit all lowercase).
+For each “no”, delete that fruit from the list.
+For any answer that is not “yes” or “no”, prompt the user to answer with one of those two values (a while loop is good here)
+Display the list.
+'''
+
+def series3():
+	fruits = series1.fruits 
+	for i in fruits:
+		print ("Do you like "+i.lower()+" ?")
```
PythonCHB, line 71: this would be a good place for an f-string:

`print(f"Do you like {i.lower()} ?")`

students/JasneetChandok/Lesson03/listlab.py
```diff
@@ -0,0 +1,100 @@
...
+		response =input('Enter y or n-->')
+		if response.lower() == 'y':
+			fruits.remove(i)
```
PythonCHB, line 74: careful here -- if you remove an item from a list while looping through it, it'll skip an item!

run this and watch the results carefully.

students/JasneetChandok/Lesson03/slicinglab.py
```diff
@@ -29,7 +29,7 @@ def elements_reversed(seq):
 
 def first_last_middle_thrid(seq):
 	'''Last third, then first third, then the middle third in the new order'''
-	new_sequence = seq[-3:][::-1] + seq[:3] + seq[(len(seq)//3):(len(seq)//3)*2]
+	new_sequence = seq[-3:] + seq[:3] + seq[(len(seq)//3):(len(seq)//3)*2]
```
PythonCHB, line 32: the first part will always be the first 3 and last 3 -- we want the first full third of the sequence!

PythonCHB (changes requested):
do take a look at the issues I brought up.

particularly the one where you remove items from a list as you loop over it...

but I'll give you credit in any case -- you've mostly got it down.

PR #80, opened 2019-04-28
Updated files and finished labs and assignments

gjkim44:
please review files


students/GKim/lesson03/list_lab.py
```diff
@@ -68,29 +70,65 @@ def main():
...
-                pass
+                print("Please answer Yes or No (y/n)")
         print_list(new_inv)
+        return False
+        
+def series_4():
+
+
+    # rev_inv = [[x[::-1]] for x in inv][::-1]
+    inv.pop(-1)
```
nairraghav, line 119: pop() will always remove the last item in the list so you don't need to specify the -1

students/GKim/lesson03/list_lab.py
```diff
@@ -68,29 +70,65 @@ def main():
...
+    print("original list: ", inv)
+    reverseWords = [reverse_word(word) for word in inv]
+    print("reversed letters list: ", reverseWords)
+
+def main():
```
nairraghav, line 124: Nice job!

gjkim44:
Please use my most recent pull request.  I had to reactivate the index that I commented out in my thank you function because I am going backwards iterating and it is to make sure that it goes to the right index.

students/GKim/lesson03/mailroom.py
```diff
@@ -0,0 +1,186 @@
...
+    clears the command screen
+    """
+    os.system("cls" if os.name == "nt" else "clear")
+
+
+def show_list():
+    """
+    shows the donors in the list without donation amounts
+    """
+    print()
```
nairraghav, line 23: To avoid the excessive print() calls, you can add a "\n" at the beginning/end of the following/previous print

ex: print("\n" +  "#" * 9, "The Current Donor List", "#" * 9)

students/GKim/lesson03/mailroom.py
```diff
@@ -0,0 +1,186 @@
...
+        thanks_answer = menu_thank_you().strip()
+        # thanks_answer = thanks_answer.strip()
+        if thanks_answer.upper() == "LIST":
+            clear_screen()
+            show_list()
+        elif thanks_answer.upper() == "MENU":
+            clear_screen()
+            break
+        else:
+            idx = len(main_donors)-1
```
nairraghav, line 98: You can remove this since you are not using it anywhere

nairraghav: Same for line 107

students/GKim/lesson03/mailroom.py
```diff
@@ -0,0 +1,186 @@
...
+            for donor in stats_list:
+                print("{:<20}${:>15} {:>15} ${:>15}".format(donor[0], donor[1], donor[2], donor[3]))
+            print("\nEnd of Report\n")    
+       
+
+def quit():
+    print("You are leaving the Mailroom!")
+    sys.exit()
+
+def mailroom_main():
```
nairraghav, line 169: Really nice job here George, you went above and beyond with some of the small additions! 

Just a note moving forward, you usually want to remove commented code when making pushes. Otherwise, it can leave your code looking messy and cluttered without a need for it to be

students/GKim/lesson03/strformat_lab.py
```diff
@@ -0,0 +1,62 @@
...
+    task_4 = f"{tup2[3]:0>2d}, {tup2[4]}, {tup2[2]}, {tup2[0]:0>2d}, {tup2[1]}"
+    print(task_4)
+
+def task_5(four_elements):
+    fourElem = f"the weight of and {four_elements[0]} is {four_elements[1]} and the weight of a {four_elements[2]} is {four_elements[3]}"
+    print(fourElem)
+
+def task_6():
+    t_six = print("{:<20}|{:^15}|{:^15}|{:>15}".format('First', '$99.01', 'Second', '$88.09'))
+    print(t_six)
```
nairraghav, line 50: Since you're calling print in the line above, you don't need to call this. 

You can just have:
def task_6():
 print("{:<20}|{:^15}|{:^15}|{:>15}".format('First', '$99.01', 'Second', '$88.09'))

nairraghav:
Great job overall George!

PR #81, opened 2019-04-29
Submission Assignment 3

archwu:
Please review:)


archwu:
String formatting lab is still on it's way

students/ArchWu/Session1/.vscode/settings.json
```diff
@@ -1,3 +0,0 @@
-{
```
PythonCHB, line 1: it would be a good idea to put a .gitignore file in your dir. then put:

.vscode/*

in it.

:-)

students/ArchWu/Session3/list_lab.py
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env python3
+fruits = ["Apples", "Pears", "Oranges", "Peaches"]
+
+def series1(fruits):
+    answer = input("Please pick a fruit to add > ")
+    fruits.append(answer)
+    print(fruits)
+    answer = input("Please pick an integer > ")
+    while not answer.isdigit() or int(answer) > len(fruits):
```
PythonCHB, line 9: nice!

students/ArchWu/Session3/list_lab.py
```diff
@@ -0,0 +1,66 @@
...
+    if answer in fruits:
+        fruits.remove(answer)
+    #bonus
+    new_fruits = fruits * 2
+    answer = input("Please pick a fruit to delete(bonus) > ")
+    while answer not in new_fruits:
+         answer = input("Please pick a fruit to delete(bonus) > ")
+    for f in fruits:
+        if f == answer:
+            new_fruits.remove(f)
```
PythonCHB, line 36: is this going to remove multiple instances of the same thing?

students/ArchWu/Session3/list_lab.py
```diff
@@ -0,0 +1,66 @@
...
+
+def series3(fruits):
+
+    for f in fruits[:]:
```
PythonCHB, line 40: perfect!

students/ArchWu/Session3/list_lab.py
```diff
@@ -0,0 +1,66 @@
...
+                break
+            else:
+                print("Please say 'yes' or 'no'.")
+    print(fruits)
+
+def series4(fruits):
+    new_fruits = []
+    for f in fruits:
+        new_fruits.append(f[::-1])
+    fruits.pop()
```
PythonCHB, line 56: why pop off one fruit here?

students/ArchWu/Session3/mailroom.py
```diff
@@ -0,0 +1,106 @@
...
+#donors [donor1, donor2]
+donors = []
+# The data struction used in this program, a list of tuples
+
+
+def add_donor(name, donation = 0):
+    """ Adds a new donor with an initial donation """
+    donations = [donation]
+    donors.append((name, donations))
+    return
```
PythonCHB, line 15: an empty return returns None. So I usually do:

return None

or just leave it off -- as None will get returned anyway.

students/ArchWu/Session3/mailroom.py
```diff
@@ -0,0 +1,106 @@
...
+    """ Adds a donation to an existing donor """
+    for donor, donations in donors:
+        if name == donor:
+            donations.append(donation)
+
+def report():
+    """ Make a report on donations received by a style as specified """
+    print("Donor Name                | Total Given | Num Gifts | Average Gift")
+    print("------------------------------------------------------------------")
+    donors.sort(reverse = True, key = lambda item: (sum(item[1])))
```
PythonCHB, line 27: went for the lambda! nice!

students/ArchWu/Session3/mailroom.py
```diff
@@ -0,0 +1,106 @@
...
+    for donor in donors:
+        name, total, num_gifts, avg = gen_stats(donor)
+        part_1 = "{}{}".format(name, ' ' * (27 - len(name)))
+        part_2 = "${}{}".format(' ' * (11 - len(str(total))), total)
+        part_3 = "{}{}{}{}{}".format(' ' *(12 - len(str(num_gifts))), num_gifts,'  $', ' ' * (12 - len(str(avg))), avg)
```
PythonCHB, line 32: you can use format specifier to make a filed be a given length:
```
 "{:10d}{:15.2f}".format(12, 1/3) 
```
better than working out the space yourself.

look it up - there are a LOT of options to control the output.

students/ArchWu/Session3/mailroom.py
```diff
@@ -0,0 +1,106 @@
...
+        result = part_1 + part_2 + part_3
+        print(result)
+    return
+
+def gen_stats(donor):
+    """ A function calculates the total and average donation of a donor """
+    name = donor[0]
+    donations = donor[1]
+    total = round(sum(donations),2)
```
PythonCHB, line 41: you can control how many digits you output --maybe be better than rounding.

students/ArchWu/Session3/mailroom.py
```diff
@@ -0,0 +1,106 @@
...
+            name = answer
+            new_user = True
+            break
+
+    while True:
+        amount = input("How much have you just donated? > ")
+        if amount == 'quit': return
+        try:
+            num_donation = float(amount)
+        except:
```
PythonCHB, line 69: nice! you're ahead of the game here -- we'll get to exception handling soon.

students/ArchWu/Session3/mailroom.py
```diff
@@ -0,0 +1,106 @@
...
+            print("Please type a float number")
+        else:
+            if new_user:
+                add_donor(name, num_donation)
+            else:
+                add_donation(name, num_donation)
```
PythonCHB, line 75: you seem to be doing the same thing, whether it's a new_user or not -- why the if?

students/ArchWu/Session3/mailroom.py
```diff
@@ -0,0 +1,106 @@
...
+            break
+    greetings = 'Dear {},\n'.format(name)
+    body = '\nThank you for your generous gift ${} to us!\n'.format(num_donation)
+    ending = '\nSincerely,\n  ABC foundations'
+    print(greetings + body + ending)
```
PythonCHB, line 80: you can make one big string with the whole thing, rather than adding the pieces together afterward.

students/ArchWu/Session3/slicing_lab.py
```diff
@@ -0,0 +1,37 @@
...
+def remove_first4_last4(seq):
+    a_new_sequence = seq[4:-4]
+    return a_new_sequence
+
+def reverse(seq):
+    a_new_sequence = seq[::-1]
+    return a_new_sequence
+
+def thirds(seq):
+    a_new_sequence = seq[2 * len(seq)//3:] + seq[:len(seq)//3] + seq[len(seq)//3:2*len(seq)//3]
```
PythonCHB, line 19: probably worth assigning a variable to `len(seq)//3` -- rather than re-computing it four times.

but otherwise, exactly right.

students/ArchWu/Session3/strformat_lab.py
```diff
@@ -0,0 +1,43 @@
...
+    data = [("First", 1, 20),("Second", 100, 20000),("Third", 10000, 200000000)]
+    for dat in data:
+        print("Name:{0:20} Age:{1:^20} cost:{2:5}".format(*dat))
+    #extra
+    tup = (1,2,3,4,5,6,7,8,9,10)
+    for elem in tup: print("{:5}".format(elem), end = '')
+
+
+
+if __name__ == '__main__':
```
PythonCHB, line 36: all good!

PythonCHB (approved):
really nice work!

students/ArchWu/Session3/mailroom.py
```diff
@@ -0,0 +1,106 @@
...
+        if answer == '1':
+            thank_you()
+        elif answer == '2':
+            report()
+        elif answer == '3':
+            sys.exit(0)
+        else:
+            print("please answer 1, 2 or 3")
+
+if __name__ == '__main__':
```
nairraghav, line 103: Nice job!

students/ArchWu/Session3/slicing_lab.py
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env python3
```
nairraghav, line 1: 👍

students/ArchWu/Session3/list_lab.py
```diff
@@ -0,0 +1,66 @@
...
+    print(fruits[int(answer) - 1])
+    answer = input("Please pick a fruit to add > ")
+    new_fruit = [answer]
+    fruits = new_fruit + fruits
+    print(fruits)
+    answer = input("Please pick a fruit to add > ")
+    fruits.insert(0, answer)
+    print(fruits)
+    for f in fruits:
+        if f[0] == 'P': print(f)
```
nairraghav, line 20: One extra case you could consider is if the user enters a fruit with a lowercase p

students/ArchWu/Session3/list_lab.py
nairraghav, line 40: You will want to be careful here when iterating through fruits. If you remove a fruit, the iterator runs into an issue and will skip a fruit. Try to address this issue. Hint: you may want to choose a different way to keep track of indices.

students/ArchWu/Session3/list_lab.py
archwu: Yes, it is the bonus objective in series 2
> (Bonus: Multiply the list times two. Keep asking until a match is found. Once found, delete all occurrences.)

students/ArchWu/Session3/list_lab.py
archwu: Yea actually during Chris's office hour he pointed out this issue and let me solve it with making a copy of fruit list by`for f in fruits[:]:` than the `for f in fruits:`

students/ArchWu/Session3/list_lab.py
archwu: Yea I should probably add a doc to explain it. I think it's included in tasks of series 4

> Delete the last item of the original list. Display the original list and the copy.

students/ArchWu/Session3/mailroom.py
archwu: During coding this part I did not start on the str_formatter, therefore I kind of hardcode the format here. Will definitely make some changes here with the formatter.

students/ArchWu/Session3/mailroom.py
archwu: I had two functions to handle the new-user situation and existing-user situation, which is at line 11 `add_donor(name, donation)` and line 17 `add_donation(name, donation)`. I think I am following the idea that one function should do exactly one thing so I coded like this. Let me know if this is good or not.

students/ArchWu/Session3/slicing_lab.py
archwu: Very reasonable, will make changes

PR #83, opened 2019-04-29
add list_lab and slicingLab

momosally:


students/Sally_Shen/lesson3/slicingLab.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+print(reversed_elements("hello world!"))
+
+
+def new_order(seq):
+    '''with the last third, then first third, then the middle third in the new order.
+    first try:
+
+    sep1 = int(len(seq) / 3)
```
PythonCHB, line 45: if you use //, you get an int.

students/Sally_Shen/lesson3/slicingLab.py
```diff
@@ -0,0 +1,88 @@
...
+    sep2 = int((len(seq) - sep1)/2)+sep1
```
PythonCHB, line 46: this can be simplified -- I think it's simply `sep1 * 2` :-)

PythonCHB (approved):
great work!

PR #84, opened 2019-04-29
Session 3 homework

Charlie-England:
completed 3 labs (list, string, and slicing) and completed mailroom part 1


students/charlie_england/session03/list_lab.py
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env python3
+print("working")
+
+#Series 1
+lst = ["Apples", "Pears", "Oranges","Peaches"]
+print(lst)
+lst.append(input("Please add another fruit"))
+print(lst)
+user_input = input("Please enter a number >>>")
+#print(f"{user_input} cooresponds to {lst[int(user_input)-1]}")
```
nairraghav, line 10: You may want to uncomment this :)

students/charlie_england/session03/mailroom.py
```diff
@@ -0,0 +1,78 @@
...
+        if name_input in donors.keys():
+            donors[name_input].append(float(input("Please enter a donation amount: >>> "))) #Ask for donation amount and append to key in dict
+            return compose_thank_you(name_input) #Call compose thank you function and return breaking the loop
+        elif name_input == "list":
+            for name in donors.keys(): print(name)
+        else:
+            #if the name_input was not list and was not already in the donors dict, will ask if the user wants to add
+            new_name_decision = input(name_input + " not found, would you like to add a new donator? y/n \n>>>").strip()
+            if new_name_decision.lower() == "y": 
+                donors.update({name_input:[input("Please enter a donation amount: >>> ")]})
```
nairraghav, line 49: Don't forget to convert this into a number, this is still being read as a string. Try making a donation with a new user and running a report, you will run into an issue.

nairraghav:
Nice job overall Charlie! I will merge this for now, please address the comment made for the mailroom input conversion.

students/charlie_england/session03/mailroom.py
Charlie-England: Thanks for catching that! I guess I never tried to run a report after I added a new user. I've addressed the issue now by adding int() in front of input. I keep forgetting that input always returns a string.

students/charlie_england/session03/list_lab.py
Charlie-England: Actually I may ask for help on this in after class. I commented it out because visual studio code kept throwing an error. When I would run it in VS code, 2 things would be "auto-inputted" 'C:/Users/Charl/Anaconda3/Scripts/activate'] and conda activate base so if I have an input that requires a number I get an error thrown. I'll go ahead and uncomment this since I'm fixing the mailroom.py code.

PR #85, opened 2019-04-29
String format lab has been completed

jchando1:


students/JasneetChandok/Lesson03/strformatlab.py
```diff
@@ -0,0 +1,90 @@
...
+
+#Part 1
+items = [("Phone", 5, 2500.00),("Charger", 7, 10.00),
+         ("Sim", 15, 9.00),("Accessories", 9, 750.00)]
+
+for i in items:
+    print("Product Name: {:15s} | Product Age: {:5d} | Product Price: {:.2f}".format(*i))
+
+#Part 2
+tup3 = range(0,10)
```
PythonCHB, line 88: careful with naming -- this isn't a tuple :-)

though it still works.

but did you need to make a big long string of the same format specifier? could you have done that with more compact code?

PR #88, opened 2019-04-30
Finished List Lab

FredBallyns:
Took a while to figure out how to make copy list different from original and not to use list being edited as index of for loop


students/FredB/session03/list_lab.py
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/env python3.7.3
```
nairraghav, line 1: Nice job!

nairraghav:
Nice job Fred!

PR #89, opened 2019-04-30
What I have so far

miotke:
I got stuck in portions of the mailroom program. Hoping we go over some things in class tomorrow :(


PythonCHB:
giving partial credit for string formatting -- and we'll wait for mailroom to be finished.

PythonCHB:
Then there is the slicing and list labs ...

PR #91, opened 2019-04-30
mailroom.py fix and list.py uncomment

Charlie-England:
fixed error with mailroom.py via a string to int conversion.

Uncommented a line in list.py


nairraghav:
Nice job! :)

PR #92, opened 2019-04-30
Add string formatting exercise

momosally:
Just finish the string formatting exercise.

### DIFF
--- a/students/Sally_Shen/lesson3/mailroom.py
+++ b/students/Sally_Shen/lesson3/mailroom.py
@@ -1,0 +1,96 @@
+
+
+'''choms'''
+
+import sys
+
+print("in mailroom")
+'''Writing first name is easier for testing.'''
+donors ={"Fred": [100, 200, 50], "Barney": [100, 50, 600], "William": [200, 50, 500],
+         "Peddle": [300, 40, 200]}
+
+
+def thank_you():
+    response = input("Type your donor name, if don't know, type list to pick")
+    '''If the user types list show them a list of the donor names and re-prompt'''
+    if response == "list":
+        for key in donors.keys():
+            print(key)
+        response = input("please type your donor name")
+
+    if response not in donors.keys():
+        donors[response] = []
+
+    newDonation = input("how much to donate?")
+    donors[response].append(newDonation)
+    email = '''
+    Dear {0},
+    Thank you for donating {1}!\n'''.format(response, newDonation)
+    print(email)
+
+
+def gen_stats(donor):
+    donations = donor[1]
+    total = sum(donations)
+    num  =  len(donations)
+    stats = (donor[0], total, num, total/num)
+    return
+
+def report():
+    amountToNames = {}
+    for name, amount in donors.items():
+        total = sum(amount)
+        if total in amountToNames.keys():
+            amountToNames[total].append(name)
+        else:
+            amountToNames[total] = [name]
+
+    sortedAmount = sorted(amountToNames, reverse=True)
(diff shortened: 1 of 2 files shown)


students/Sally_Shen/lesson3/mailroom.py
```diff
@@ -0,0 +1,96 @@
...
+    '''If the user types list show them a list of the donor names and re-prompt'''
+    if response == "list":
+        for key in donors.keys():
+            print(key)
+        response = input("please type your donor name")
+
+    if response not in donors.keys():
+        donors[response] = []
+
+    newDonation = input("how much to donate?")
```
nairraghav, line 24: Don't forget that once you take the input of the donation amount, you need to convert it to an integer. Otherwise it reads in as a string and the next line throws an error!

students/Sally_Shen/lesson3/strformat_lab.py
```diff
@@ -0,0 +1,121 @@
+'''
+Task One
```
nairraghav, line 2: Nice job here!

nairraghav:
Please address the input int issue and I will merge it in! As a note, please do try to run your code to make sure it is fully functional. It is good to try various scenarios to make sure that your code can handle it. Specifically, if you ran 1 or 2 by itself, they would work. But if you add a new user and then try to create a report, you will run into an error.

students/Sally_Shen/lesson3/mailroom.py
```diff
@@ -0,0 +1,96 @@
...
+    if response == "list":
+        for key in donors.keys():
+            print(key)
+        response = input("Please type your donor name.")
+
+    if response not in donors.keys():
+        donors[response] = []
+
+    newDonation = input("how much to donate?")
+    donors[response].append(float(newDonation))
```
PythonCHB, line 25: so you do turn it into a number here -- good.

but then you use the string version is letter below.

you really want to make it a number first, and use the number both when adding to the donations list and in the letter - that way you can use string formatting in the letter to present it correctly: $200.00, for instance.

students/Sally_Shen/lesson3/mailroom.py
```diff
@@ -0,0 +1,96 @@
...
+    total = sum(donations)
+    num  =  len(donations)
+    stats = (donor[0], total, num, total/num)
+    return
+
+def report():
+    amountToNames = {}
+    for name, amount in donors.items():
+        total = sum(amount)
+        if total in amountToNames.keys():
```
PythonCHB, line 43: you are building this from scratch -- it won't alrady be there. And if it is, it's coincidence -- what if two donor happen to have the same total?

Ahh -- I see what you are doing here -- you are building a lit of names with the same total, so that you can capture duplicates. That can work, yes, but it's getting a bit too complicated.

Really better to create a list of al the donor information, and custom sort that.

students/Sally_Shen/lesson3/mailroom.py
```diff
@@ -0,0 +1,96 @@
...
+        else:
+            amountToNames[total] = [name]
+
+    sortedAmount = sorted(amountToNames, reverse=True)
+    output = "{:<15}${:<15}{:<12}${:<12}\n".format("Donor Name", "|Total Given", "|Num Gifts", "|Average Gift")
+    output += "-------------------------------------------------------------\n"
+
+
+    for amount in sortedAmount:
+        nameList = amountToNames[amount]
```
PythonCHB, line 54: Interesting -- I wouldn't have thought to do this way. But it should work, but a few issues:

There is a lot of coupling here -- a lot of this code will need to change if you change how the donors are stored. 

Better to extract all the donor info in one chunk, and work with that data structure.

PythonCHB (approved):
I'm going to merge this -- that way it will be clear what's new when you do a new PR.

PR #93, opened 2019-04-30
Adds mailroom.py

jamessward:


students/jsward/session03/mailroom.py
```diff
@@ -0,0 +1,129 @@
...
+            w4=' ' * (11 - len(average_str)),
+            a=average_str
+        ))
+
+
+def donor_management():
+    user_input = menu_donor()
+    if 'l' in user_input:
+        list_donors()
+    elif 'a' in user_input:
```
nairraghav, line 111: Interesting choice to split up adding a donor and adding a donation! Looks good 😄

PR #95, opened 2019-04-30
Mailroom Part1 w/ Python clear screen

FredBallyns:
Finished session3 assignments. Made function to clear python screen that adjusts clear screen command depending on operating system.

### DIFF
--- a/students/FredB/session03/mailroom.py
+++ b/students/FredB/session03/mailroom.py
@@ -1,0 +1,101 @@
+#Mailroom Part1
+#FredBallyns
+#Session03
+
+import sys
+import os
+
+donors = [("Fred Flintstone", [100, 200, 50]),
+        ("James Bond", [0,0,7]),
+        ("Lex Luthor", [5000,]),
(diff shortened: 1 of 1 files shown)


students/FredB/session03/mailroom.py
```diff
@@ -0,0 +1,101 @@
...
+#FredBallyns
+#Session03
+
+import sys
+import os
+
+donors = [("Fred Flintstone", [100, 200, 50]),
+        ("James Bond", [0,0,7]),
+        ("Lex Luthor", [5000,]),
+        ("Harambe", [3, 2, 1]),
```
nairraghav, line 11: RIP

students/FredB/session03/mailroom.py
```diff
@@ -0,0 +1,101 @@
+#Mailroom Part1
```
nairraghav, line 1: Nice job Fred, this looks good! Good usage of the os module to clear the screen, keeps it very neat!

PR #97, opened 2019-05-01
Session3

pberhe:


students/PeterB/Session3/list_lab.py
```diff
@@ -0,0 +1,63 @@
+#series 1
+
+Fruits = ['Apples', 'Pears', 'Oranges', 'Peaches']
+print('Fruits')
```
nairraghav, line 4: I think you meant this without the quotation marks :)

students/PeterB/Session3/list_lab.py
```diff
@@ -0,0 +1,63 @@
...
+
+New_Fruit = input('Banana')
+Fruits.append(New_Fruit)
+print(Fruits)
+
+Number = input('Pick a number')
+print(f'the {Fruits} in the list is [int(Fruits)-1]')
```
nairraghav, line 11: This doesn't print out what were expecting here. You want to print out the i-th item given in the list, when given i+1. So for example, if the user returns 4, you want to print fruits[3].

students/PeterB/Session3/list_lab.py
```diff
@@ -0,0 +1,63 @@
...
+
+Fruits = ['Bananas']+Fruits
+print(Fruits)
+
+print('Adding another fruit to the beginning of the list')
+Fruits.insert(0, 'Strawberry')
+print(Fruits)
+
+for value in Fruits:
+    if 'P' in value[0]:
```
nairraghav, line 21: You will want to be careful here because it might not always be capital P!

students/PeterB/Session3/list_lab.py
```diff
@@ -0,0 +1,63 @@
...
+        print(value)
+
+
+#series 2
+
+print(Fruits)
+del Fruits(0)
+response = input(" name a fruit to delete ")
+if response in Fruits:
+    Fruit.remove(response3)
```
nairraghav, line 31: Your list name is Fruits, this will throw an error

students/PeterB/Session3/list_lab.py
```diff
@@ -0,0 +1,63 @@
...
+else:
+    print response, "cannot be deleted because it is not found in list"
+
+
+#series 3
+print(Fruits)
+
+Fruits.lower()
+for items in Fruits:
+    Fruits.lowerappend(Fruits.lower())
```
nairraghav, line 42: I'm not sure I understand what you are aiming to do here

students/PeterB/Session3/list_lab.py
```diff
@@ -0,0 +1,63 @@
...
+
+for item in Fruits:
+    answer = ''
+    while answer ='no' and answer ='yes':
+        answer = input('Do you like '+item+'? yes/no:')
```
nairraghav, line 47: A neater way to do this would be with the f string that we learned in class

students/PeterB/Session3/list_lab.py
```diff
@@ -0,0 +1,63 @@
...
+#series 3
+print(Fruits)
+
+Fruits.lower()
+for items in Fruits:
+    Fruits.lowerappend(Fruits.lower())
+
+for item in Fruits:
+    answer = ''
+    while answer ='no' and answer ='yes':
```
nairraghav, line 46: You will never enter this while loop because the variable "answer" will never be both yes and no.

students/PeterB/Session3/list_lab.py
```diff
@@ -0,0 +1,63 @@
...
+        answer = input('Do you like '+item+'? yes/no:')
+        if answer =='no':
+            Fruits.remove(item)
+print(Fruits_3)
```
nairraghav, line 50: This doesn't exist, I believe you mean Fruits

students/PeterB/Session3/slicing.py
```diff
@@ -0,0 +1,36 @@
+# slicing
+# first and last
+seq = '3, 2, 1'
+def exchange_first_last(seq):
+    first = seq[-1]
+    mid = seq[1:-1]
+    last = seq[0]
+    return print([first] + mid + [last])
```
nairraghav, line 8: You don't need to add the brackets, this will make it a list which you don't need. You can just have first + mid  + last

students/PeterB/Session3/slicing.py
```diff
@@ -0,0 +1,36 @@
...
+my_list = "1, 2, 3, 4, 5, 6"
+def remove_every_other(my_list):
+    return my_list[::2]
+
+# first four last four
+
+def first_last_four(sequence):
+    first_four = sequence[:4]
+    last_four = sequence[-4:]
+    new_sequence = first_four + last_four
```
nairraghav, line 21: The goal of this is: with the first 4 and the last 4 items removed, and then every other item in the remaining sequence.

You want to remove the ones you have shown here and then every other item in the middle.

students/PeterB/Session3/slicing.py
```diff
@@ -0,0 +1,36 @@
...
+
+def elements_reversed(sequence):
+    reverse_sequence = sequence[::-1]
+    return reverse_sequence
+
+
+# in thirds
+
+def thirds(seq):
+    result = seq[6:] + seq[:3] + seq[3:-3]
```
nairraghav, line 34: You can't always hardcode these values because you want the logic to apply to any string not just the specific sequence you have. Hint: Use division to divide your string into thirds and go from there.

students/PeterB/Session3/mailboxmoney0.py
```diff
@@ -0,0 +1,101 @@
...
+            print(dash)
+        else:
+            print('{:20} ${:>10.1f}{:>20d} ${:>16.1f}'.format(sortedDonorSummary[i][0], sortedDonorSummary[i][1], sortedDonorSummary[i][2],  sortedDonorSummary[i][3]))
+
+
+
+
+def exitProgram():
+    print("Thank you, good bye!")
+    sys.exit()
```
nairraghav, line 81: You will need to `import sys` to run this.

nairraghav:
Your mailroom looks good but the slicing lab and list lab need some work. Make sure you run your code before you are ready to submit, there are some issues that you would run into when running this code.

PythonCHB (approved):
looking good.

PR #99, opened 2019-05-01
adding mailroom, part 1...in work + updated .gitignore file and uploaded content

BBrumleve:
Was working with Ron on my [thank_you()] and was having behavioral issues after Ron’s help. Was able to finally correct the issues just before class yesterday’s and am now uploaded the first cut. The issues I was having were: the IF stmts were inside a single for loop. This has the code run through all options and print out all options. Corrected in this submittal!


students/BrianB/.gitignore
```diff
@@ -0,0 +1,7 @@
+lesson01/monkey_trouble.py
+lesson01/sleep_in.py
+lesson03/mailroom.py
```
PythonCHB, line 3: you sure you wan to ignore this one?

students/BrianB/lesson03/mailroom1.py
```diff
@@ -0,0 +1,154 @@
...
+#  donor[1].pop()  Leaving pop() blank, deletes
+#                       the value in the last
+#                       position
+#----------------------------------------------
+
+import sys
+import math
+from textwrap import dedent
+
+print("In the Mailroom")
```
PythonCHB, line 22: this should probably be in the "if __main__" block

students/BrianB/lesson03/mailroom1.py
```diff
@@ -0,0 +1,154 @@
...
+def thank_you():
+    while True:
+        name = input("Which Donor would you like to see?\n\
+    Please enter the full name only.\n\
+    >>> ")
+        if name == "list":
+            for donor in donor_data:
+                print(donor[0])
+            continue
+        for donor in donor_data:
```
PythonCHB, line 46: the logic is fine here, but too much embedded in the while loop.

Ideally, you call out to other functions for the actual work.

students/BrianB/lesson03/mailroom1.py
```diff
@@ -0,0 +1,154 @@
...
+                                                            employer
+                                '''.format(donor[0], donor[1][-1])))
+                break  # this does not break out of [thank =_you()]
+        else:
+            print(name, ": There's a new donor!")
+            amt = input("New donor found! Please enter a donation amount:")
+            donor_data.append((name, [float(amt)]))
+            print(donor_data)
+            # print(gen_letter(donor))
+            print(dedent('''
```
PythonCHB, line 71: you defiantly don't want to repeat this code! That's EXACLY what functions are for!

generate the letter in a function, and just print it from here.

students/BrianB/lesson03/mailroom1.py
```diff
@@ -0,0 +1,154 @@
...
+
+def gen_stats(donor):
+    donations = donor[1]
+    total = sum(donations)
+    num = len(donations)
+    stats = (donor[0], total, num, total/num)
+    return stats
+
+
+def gen_letter():
```
PythonCHB, line 94: here it is! why isn't it called from above??

students/BrianB/lesson03/mailroom1.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+if __name__ == "__main__":
+    print("Welcome to the Mailroom!")
+
+    main_menu()
+
+    donor = ("Barney Rubble", [101, 201, 51])
+    assert gen_stats(donor) == ("Barney Rubble", 353, 117.67)
+
```
PythonCHB, line 154: looks like you still need the report part done. 

You can add that when you add dicts ...

students/BrianB/lesson04/dic_lab.py
```diff
@@ -0,0 +1,28 @@
+#! bin/usr/env python3
```
PythonCHB, line 1: this should be:

#!/usr/bin/env python3

but since you clearly aren't using it -- you can just leave it off.

PythonCHB (changes requested):
I"m giving you 10/15 for mailroom -- we'll bump that up when you add the reports.

BBrumleve:
Thanks for your comments Chris. It’s a work in progress. I had the separate
thank you note function but it wasn’t working calling the parameter names I
have set in the while loop. I’ll fix it just need to figure out how to link
up my while loop parameters.
BrianB

PR #101, opened 2019-05-03
session3 files

trayffs:
Mailroom is a bit rough. I cannot seem to get certain features working.

The assignment felt a bit above my head and I'm struggling. Ultimately I had to browse other people's assignments to cobble this together. I'm having a hard time conceptualizing what the right tools are for specific purposes. When to use an if/else or a while, how to select a list withing in a tuple within a list, making sure I don't get "unsupported operand types for int and str", etc.

I can't spend any more time on this as I'd like to try to put a dictionary version together and get cracking on session4.

### DIFF
--- a/students/TylerRay/session3/mailroom.py
+++ b/students/TylerRay/session3/mailroom.py
@@ -1,0 +1,92 @@
+#!/usr/bin/env python3
+
+
+donors = [("Alpha Beta", [1000.21, 250.80]),
+          ("Tucker Jones", [800.33]),
+          ("Vladmir Putin", [500000.12, 250000.55, 750000]),
+          ("Kim Jong Il", [200.80]),
+          ("Tony Robinson", [500000, 1000000, 750000])]
+
+donor = donors[0]
+donations = donors[1]
+
+def main_menu():
+    '''
+    Main menu for the program.
+    Will loop through and requests input until 3 is entered and then quits.
+    1 will allow an addition of a donator, a new donation and the printing of a thank you letter
+    2 will print a report of the donators and their donations as well as their average donation.
+    '''
+    while True:
+        choice = input("""What would you like to do?
+            1: Send a thank you
+            2: Create a Report
+            3: Quit
+            >>>""").strip()
+        if choice == "1":
+            # leads to the thank_you function
+            thank_you()
+        elif choice == "2":
+            # leads to the report function
+            for line in report():
+                print(line)
+        elif choice == "3":
+            # breaks the loop effectively exiting the program
+            print("Quitting...")
+            break
+        else:
+            # if the item is not 1, 2, or 3, will print this and go back to beginning of while loop
+            print(("You replied {}, please reply with 1, 2 or 3").format(answer))
+
+
+def thank_you():
+    donor_input = input("Enter full name of donor else type 'list' to see donor list: ")
+    if donor_input == 'list':
+        for donor in donors:
+            print(donor)
+
+
+    else:
+        # Search if repeat donor
+        exists = False
+        for donor in donors:
+            if donor[0] == donor_input:
+                donor_input = donor[0]
+                exists = True
+        if exists:
+            pass
+        else:
+            donors.append((donor_input, []))
+
+        # Add new donation
+        donation = int(input("Enter donation amount: "))
+        for donor in donors:
+            if donor[0] == donor_input:
+                donor[1].append(donation)
+
+        print("Esteemed Donors", '\n', '------------')
+        for donor in donors:
+            print(donor)
+        print (f"Thank you {donor_input} for your generous donation of ${donation} from a charity")
+
+
+def report():
+    """
+    return a report by adding a line for every name in the donor dictionary
+    """
+    report_list = ["Donor Name" + " " * 10 + "| Total Given " + "| Num Gifts " + "| Average Gift\n" + "-" * 60]
+
+    for donor in donors:
+        stats = []
+        donations = donors[1]
+        total = sum(donations)
+        num = len(donations)
+        stats.append((donor[0], total, num. total / num))
+        print(stats)
+
+    return donations
+
+
+if __name__ == "__main__":
+    print("Welcome to the Mailroom!")
+    main_menu()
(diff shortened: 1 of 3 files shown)


PythonCHB:
Sorry it's being a struggle -- but that's why you're doing it :-) -- no way to learn coding other than to write code.

for the unsupported operand errors -- take the time to think at every step what type you variables are:

input()will always return a string -- if it needs to be a number, then you have to make it so -- with int() or float()

for is for lopign through a list of items, or looping a known number of time (for i in range(x)) -- whiel is for when you need to do something over an over again, but you can't know how many times.

students/TylerRay/session3/mailroom.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+donors = [("Alpha Beta", [1000.21, 250.80]),
+          ("Tucker Jones", [800.33]),
+          ("Vladmir Putin", [500000.12, 250000.55, 750000]),
+          ("Kim Jong Il", [200.80]),
+          ("Tony Robinson", [500000, 1000000, 750000])]
+
+donor = donors[0]
+donations = donors[1]
```
PythonCHB, line 11: ????

I suspect you aren't actually using these...

students/TylerRay/session3/mailroom.py
```diff
@@ -0,0 +1,92 @@
...
+            1: Send a thank you
+            2: Create a Report
+            3: Quit
+            >>>""").strip()
+        if choice == "1":
+            # leads to the thank_you function
+            thank_you()
+        elif choice == "2":
+            # leads to the report function
+            for line in report():
```
PythonCHB, line 31: probably a little better to have this loop in the report() function.

students/TylerRay/session3/mailroom.py
```diff
@@ -0,0 +1,92 @@
...
+                print(line)
+        elif choice == "3":
+            # breaks the loop effectively exiting the program
+            print("Quitting...")
+            break
+        else:
+            # if the item is not 1, 2, or 3, will print this and go back to beginning of while loop
+            print(("You replied {}, please reply with 1, 2 or 3").format(answer))
```
PythonCHB, line 39: nice

students/TylerRay/session3/mailroom.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def thank_you():
+    donor_input = input("Enter full name of donor else type 'list' to see donor list: ")
+    if donor_input == 'list':
+        for donor in donors:
+            print(donor)
```
PythonCHB, line 46: you should only print the name here. if you had a generaous donor for years, it could have a long list of donations...

students/TylerRay/session3/mailroom.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+    else:
+        # Search if repeat donor
+        exists = False
+        for donor in donors:
+            if donor[0] == donor_input:
+                donor_input = donor[0]
```
PythonCHB, line 54: these are already equal -- why reset it?

students/TylerRay/session3/mailroom.py
```diff
@@ -0,0 +1,92 @@
...
+    donor_input = input("Enter full name of donor else type 'list' to see donor list: ")
+    if donor_input == 'list':
+        for donor in donors:
+            print(donor)
+
+
+    else:
+        # Search if repeat donor
+        exists = False
+        for donor in donors:
```
PythonCHB, line 52: this could be a bit simpler, by using break:

for donor in donors:
    if donor[0] == donor_input:
        break
    else:
        donor = None


and "donor" is the donor object you found.

`else` on an if block is kinda weird, but it is handy.

students/TylerRay/session3/mailroom.py
```diff
@@ -0,0 +1,92 @@
...
+        else:
+            donors.append((donor_input, []))
+
+        # Add new donation
+        donation = int(input("Enter donation amount: "))
+        for donor in donors:
+            if donor[0] == donor_input:
+                donor[1].append(donation)
+
+        print("Esteemed Donors", '\n', '------------')
```
PythonCHB, line 67: what's this for?

students/TylerRay/session3/mailroom.py
```diff
@@ -0,0 +1,92 @@
...
+                exists = True
+        if exists:
+            pass
+        else:
+            donors.append((donor_input, []))
+
+        # Add new donation
+        donation = int(input("Enter donation amount: "))
+        for donor in donors:
+            if donor[0] == donor_input:
```
PythonCHB, line 64: you already did this search, see if you can use the donor you found before.

students/TylerRay/session3/mailroom.py
```diff
@@ -0,0 +1,92 @@
...
+    """
+    report_list = ["Donor Name" + " " * 10 + "| Total Given " + "| Num Gifts " + "| Average Gift\n" + "-" * 60]
+
+    for donor in donors:
+        stats = []
+        donations = donors[1]
+        total = sum(donations)
+        num = len(donations)
+        stats.append((donor[0], total, num. total / num))
+        print(stats)
```
PythonCHB, line 85: this is going to keep printing teh whole list as it builds up -- not what you want.

and you need to use some string formatting to get a nice table.

students/TylerRay/session3/strformat_lab.py
```diff
@@ -0,0 +1,17 @@
+tpl1 = ( 2, 123.4567, 10000, 12345.67)
+
+def printTuple(tpl1):
+    v1 = "file_{0:03}: ".format(tpl1[0])
+    v2 = "{0:.2f}".format(tpl1[1])
+    v3 = "{:.2e}".format(tpl1[2])
+    v4 = "{0:.2e}".format(tpl1[3])
+    return f"{v1} {v2} {v3} {v4}"
```
PythonCHB, line 8: you should be able to do this in one big format string.

PythonCHB (approved):
I'm merging and giving you credit for mailroom -- you can continue to clean it up as you add to it in the next couple assignments.

PR #102, opened 2019-05-04
adding lesson04 FINAL: dict_lab.py, trigrams.py; IN WORK: file_ext.py

BBrumleve:
approx 95% complete

### DIFF
--- a/students/BrianB/lesson04/dic_lab.py
+++ b/students/BrianB/lesson04/dic_lab.py
@@ -1,36 +1,15 @@
 #! bin/usr/env python3
 
-'''
-Dictionaries 1
-    Create a dictionary containing “name”, “city”, and “cake” for
-        “Chris” from “Seattle” who likes “Chocolate” (so the keys
-        should be: “name”, etc, and values: “Chris”, etc.)
-        
-    Display the dictionary.
-    Delete the entry for “cake”.
-    Display the dictionary.
-    Add an entry for “fruit” with “Mango” and display the dictionary.
-        Display the dictionary keys.
-        Display the dictionary values.
-        Display whether or not “cake” is a key in the dictionary
-            (i.e. False) (now).
-        Display whether or not “Mango” is a value in the dictionary
-            (i.e. True).
-
-Dictionaries 2
-    Using the dictionary from item 1: Make a dictionary using
-        the same keys but with the number of ‘t’s in each value
-        as the value (consider upper and lower case?).
-
-'''
+d = {
+    'name': 'Chris',
+    'city': 'Seattle',
+    'cake': 'Chocolate'
+    }
 
 
 def dict_1():
-    d1 = {
-        'name': 'Chris',
-        'city': 'Seattle',
-        'cake': 'Chocolate'
-        }
+    """Working with a dictionary"""
+    d1 = d
 
     # display dictionary
     print('Original Dictionary: ', d1)
@@ -46,7 +25,7 @@ def dict_1():
 
     # display dictionary with new key and value
     print('Key, value with fruit, mango added:')
-    for (key, value) in d1.items():
+    for key, value in d1.items():
         print(key, ': ', value)
 
     # display the dictionary keys
@@ -64,20 +43,22 @@ def dict_1():
     for key, value in d1.items():
         if value == 'Mango':
             print('\nIs mango a value?', True, '\n')
-
-    # print number of t's in values
-    # expect to see {name: 0, city: 2, fruit: 0}
-    for key, value in d1.items():
-        for item in value:
-            if 't' in item.lower():
-                print(key, ':', len(item))
-            else:
-                print(key, ':', 0)
-            break
     return d1
+
+
+def dict_2():
+    """Working with a dictionary"""
+    d2 = d
+    print("The count of 't's in each value are: ")
+    while True:
+        for key, value in d2.items():
+            print(key, ':', value.lower().count('t'))
+        break
+    return d2
 
 
 # display dictionary keys:
 dict_1()
 
-
+# dictionary 2 function call
+dict_2()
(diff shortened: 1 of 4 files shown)


students/BrianB/lesson04/dic_lab.py
```diff
@@ -1,28 +1,64 @@
...
+
+    # display dictionary without cake
+    print('\nDictionary with key "cake" deleted: ', d1, '\n')
+
+    # add dictionary with a fruit
+    d1.update({'fruit': 'Mango'})
+
+    # display dictionary with new key and value
+    print('Key, value with fruit, mango added:')
+    for key, value in d1.items():
```
nairraghav, line 28: Another way to do this would be:
for item in d1:
    print(f'{item} : {d1[item]}"

students/BrianB/lesson04/dic_lab.py
```diff
@@ -1,28 +1,64 @@
...
+        print(key, ': ', value)
+
+    # display the dictionary keys
+    print('\n', d1.keys())
+
+    # display the dictionary values
+    print('\n', d1.values())
+
+    # test if 'cake' is no longer a key
+    for key, value in d1.items():
```
nairraghav, line 38: Here you're not actually validating that cake is no longer a key. You're checking to see if the key 'fruit' exists. What you want to do is check the key doesn't exist. There is a very simple way to do this:
if 'cake' in d1: (this will check the keys in d1 to see if "cake" exists).

students/BrianB/lesson04/dic_lab.py
```diff
@@ -1,28 +1,64 @@
...
+        if value == 'Mango':
+            print('\nIs mango a value?', True, '\n')
+    return d1
+
+
+def dict_2():
+    """Working with a dictionary"""
+    d2 = d
+    print("The count of 't's in each value are: ")
+    while True:
```
nairraghav, line 53: You actually don't need this while loop here. Just iterating through the dictionary is good

students/BrianB/lesson04/dic_lab.py
```diff
@@ -1,28 +1,64 @@
...
+        for key, value in d2.items():
+            print(key, ':', value.lower().count('t'))
```
nairraghav, line 55: The goal isn't to just print the value here, you want to actually update the value of d2 to the number.

students/BrianB/lesson04/dic_lab.py
```diff
@@ -1,28 +1,64 @@
...
+        break
+    return d2
+
+
+# display dictionary keys:
+dict_1()
+
+# dictionary 2 function call
+dict_2()
```
nairraghav, line 64: I believe you are still missing the set1 and set1 exercises

students/BrianB/lesson04/file_ext.py
```diff
@@ -1,3 +1,92 @@
...
+lines have nicknames, and other small differences, so you will need to write 
+some code to make sure you get it all correct.
+
+Extra challenge: keep track of how many students specified each language.
+'''
+
+#import os
+#import glob
+
+#def file_path():
```
nairraghav, line 58: Very good! The exercise asks for all files so you don't need to specify just python files. Also by having just 1 * is enough to act as the wildcard.

students/BrianB/lesson04/file_ext.py
```diff
@@ -1,3 +1,92 @@
...
+#    """Finds python files and Returns their directory."""
+#
+#    for item in glob.glob('**.py'):
+#        print(os.path.abspath(item))
+#    return
+
+
+#def src_dst():
```
nairraghav, line 66: There is a simpler way to do this without using the cp command. Refer to the instructions, you should be able to do this with the open function.

ex: open(filename, 'rb') / open(filename, 'wb')

students/BrianB/lesson04/file_ext.py
```diff
@@ -1,3 +1,92 @@
...
+#    return
+
+
+def prog_langs():
+    """Reads "students.txt" and Returns programming
+        languages known by the students."""
+
+    for line in open('students.txt'):
+        for language in line.split(": "):
+            if language == language.lower():
```
nairraghav, line 80: After doing the split on ": ", your language variable becomes:  Mick, shell, python. Here you want to split again to get the names Mick, shell, python and then do your lower check to validate if it's a language. Otherwise you're always checking the very first item which will always fail for lines nicknames

students/BrianB/lesson04/file_ext.py
```diff
@@ -1,3 +1,92 @@
...
+                print([language])
```
nairraghav, line 81: We don't want to print the language right away. You will want to store them all and print it at the very end, otherwise you'll print out duplicates. What kind of data structure should you use here if the goal is to not have duplicates?

**Bonus** If you wanted to go a step further and keep track of how many times each language was shown, what data structure would you want to use?

students/BrianB/lesson04/trigrams.py
```diff
@@ -1,50 +1,44 @@
...
-BrianB/lesson04/Trigram text samples/sherlock_small.txt")
-# print(words)
-trigrams = build_trigrams(words)
-print(trigrams)
+    BrianB/lesson04/sherlock_small.txt")
+
+
+if __name__ == "__main__":
+
+    print_trigram()
```
nairraghav, line 44: This is really good! The logic is pretty sound here, you will just want to be careful of punctuation which gets in the way of your trigrams. For example ("I,", "I") is different from ("I", "I"). You can try to combat this by using the string module.

import string
string.punctuation #'!"#$%&\'()*+,-./:;<=>?@[\\]^_`{|}~'

Using this, you can check to see if the last character in each word is punctuation and remove it if necessary. Also, you should use lower to get everything in the same case. These are just extra things to note but I would say this works functionally.

Do try to implement the last part of the exercise to generate sentences from your trigrams using rand.

nairraghav:
There are quite a few notes in here, please address them in a future PR. Overall, the majority of this is _mostly_ there, nice job!

PR #106, opened 2019-05-05
Adding session04 files

ervbrian:
Included files:
students/BrianE/session04/dict_lab.py
students/BrianE/session04/trigrams.py
students/BrianE/session04/MailRoom.py


students/BrianE/session04/MailRoom.py
```diff
@@ -26,13 +26,14 @@ def send_thankyou(donors):
             else:
                 donors[name] = [amount]
             email_message = f"Dear {name.strip()},\n" \
-                            f"    Thank you for your recent donation of ${amount}!\n\n" \
-                            f"    Sincerely,\n\n" \
-                            f"    FakeCorp, Inc"
+                f"    Thank you for your recent donation of ${amount}!\n\n" \
```
PythonCHB, line 29: if you want to put strings on multiple lines, you don't need an f before each one. And you ceratinly don't need an f for a string with no formatting in it. And you son't need the \ is it's in a parentheses. So this can be:

```
email_message = (f"Dear {name.strip()},\n" 
                  "    Thank you for your recent donation of ${amount}!\n\n" 
                  "    Sincerely,\n\n" 
                  "    FakeCorp, Inc")

File: students/BrianE/session04/MailRoom.py [hunk 2]
```diff
@@ -0,0 +1,118 @@
...
+    """
+    Generate thank-you emails to all donors in database
+    :param donors: dictionary of donors
+    :return: write email text to file for each donor
+    """
+
+    for donor in donors.keys():
+        total_donated = round(sum(donors[donor]), 2)
+        most_recent_donation = donors[donor][0]
+        email_message = f"Dear {donor},\nThank you for your recent donation of ${most_recent_donation}! In total, " \
```
PythonCHB, line 47: DRY! 

isn't this same code above? put that in a function, and then use the same function to print a letter and to write it to a file.

students/BrianE/session04/MailRoom.py
```diff
@@ -0,0 +1,118 @@
...
+            f"you've donated {total_donated}!\n\n"
+
+        with tempfile.TemporaryFile(mode='w+') as email_file:
```
PythonCHB, line 50: not very useful to write these to a tempfile ---- but nice find!

students/BrianE/session04/MailRoom.py
```diff
@@ -0,0 +1,118 @@
...
+    Generate donation report
+    :param donors: dictionary containing donor info
+    :return: None
+    """
+
+    header = f'| {"Donor Name":19} | {"Total Donated":9} | {"Num Gifts":5} | {"Average Gift":8}'
+    print(header)
+    print("-" * 64)
+
+    for donor in donors.keys():
```
PythonCHB, line 67: if you need to work with both the keys and the items, you can do that this way:

```
for name, donations in donors.items():
    ....
```

students/BrianE/session04/MailRoom.py
```diff
@@ -0,0 +1,118 @@
...
+        What would you like to do?
+        1) Send a Thank You
+        2) Create a Report
+        3) Generate thank-you emails for all donors
+        9) QUIT\n
+        >>>:"""))
+
+        try:
+            switch_main_menu.get(choice)(donors)
+        except TypeError:
```
PythonCHB, line 103: this works, 'cause dict.get returns a None if the key isn't there.

But better to not use get, an catch the KeyError instead:

```
try:
    switch_main_menu[choice](donors)
except KeyError:
    print("Please enter a valid option. (1, 2, 3 or 9)")
```

students/BrianE/session04/trigrams.py
```diff
@@ -0,0 +1,120 @@
...
+def import_text(file):
+    """
+    Import text from file and return a list of words.
+    :param file: input file containing starting text
+    :return: list of words
+    """
+
+    # Only use alpha-numeric words from file
+    with open(file=file, mode='r') as text:
+        word_list = [word for word in text.read().split() if word.isalnum()]
```
PythonCHB, line 18: what is this going to do for contractions: can't or parentheses: (this is in parens) (would "think" and "parens" get used?

students/BrianE/session04/trigrams.py
```diff
@@ -0,0 +1,120 @@
...
+    # Start of sentence
+    sentence = []
+    starting_key = random.choice(list(trigram.keys()))
+    sentence.extend([word for word in starting_key])
+    sentence.append(trigram[starting_key][random.choice(range(0, len(trigram[starting_key])))])
+
+    # Append words to sentence until length is 15 words or KeyError occurs.
+    while len(sentence) < 15:
+        try:
+            key1, key2 = sentence[-2:]
```
PythonCHB, line 65: why the names key1 and key2?

students/BrianE/session04/trigrams.py
```diff
@@ -0,0 +1,120 @@
...
+            sentence.append(trigram[(key1, key2)][random.choice(range(0, len(trigram[(key1, key2)])))])
```
PythonCHB, line 66: random.choice will make a choice from a list -- no need for range, or len, or ...

students/BrianE/session04/trigrams.py
```diff
@@ -0,0 +1,120 @@
...
+            break
+
+    # Cleanup sentence punctuation
+    sentence[0] = sentence[0].capitalize()
+    if sentence[-1][-1] != '.':
+        sentence[-1] = f'{sentence[-1]}.'
+    return sentence
+
+
+def generate_paragraph(trigram):
```
PythonCHB, line 77: I like this. nice!

students/BrianE/session04/trigrams.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+def generate_new_book(text):
+    """
+    Iterate through generated text lists display new book
+    :param text: list of sentences
+    :return: None
+    """
+
+    for paragraph in text:
```
PythonCHB, line 97: could you make this faster an more efficient with `" ".join()`

PythonCHB (approved):
nice work -- you've got dicts pretty much down :-)

PR #109, opened 2019-05-06
Lesson 4 completions

Charlie-England:
Hey I was down with a bad cold all last week so I wasnt able to really get into these lessons as much as I wish I could have but I will update them now that I'm feeling better over the next few days.

Dict and set is completed
Mailroom part 2: the file part is done but I was unable to figure out dictionaries for case swapping but I will work on that later tonight and hopefully we can touch on it in class

Trigrams: completed basic parts but much more can be done with formatting and prepping the text to be used in regards to nouns/punctuation


Charlie-England:
Just finished switch-case for mailroom.py. I could not figure out how to put an end condition in for the switch-case where it would break out of the loop.

students/charlie_england/session04/dict_lab.py
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env python3
+
+#Dictionaries 1:
+first_dict = {"name":"Chris","city":"Seattle","cake":"Chocolate"}
+print(first_dict)
+del first_dict["cake"]
+print(first_dict)
+first_dict.update({"fruit":"Mango"})
```
nairraghav, line 8: you can also do: first_dict['fruit'] = 'Mango'

students/charlie_england/session04/dict_lab.py
```diff
@@ -0,0 +1,57 @@
...
+print(first_dict.keys())
+print(first_dict.values())
+if "cake" not in first_dict.keys():
```
nairraghav, line 11: When searching keys in a dict, you don't need to specify the dict.keys()
ex: if "cake" not in first_dict 
This will return the same logic

students/charlie_england/session04/mailroom_part2.py
```diff
@@ -0,0 +1,114 @@
+#!/usr/bin/env python3
+#refactor to have the donors dict be 2 lists
+import sys
+# Donors dict Name:[[list of donations],[sum donations, num donations, average donations]]
+donors = {"Chris Christly": [[1000.21, 250.80],[]], "Bob Barley": [[800.33],[]], "Nick Nilly": [[
```
nairraghav, line 5: Be careful here, you don't actually want to have a list of lists for the donations. You can just have something like:
{{"Chris Christly": [1000.21, 250.80], "Bob Barley": [800.33], ... ] 

Otherwise, you will run into issues when trying to compute stats. When I add more donations, it doesn't update stats properly and I believe this is the issue.

students/charlie_england/session04/trigrams.py
```diff
@@ -0,0 +1,80 @@
...
+"""
+
+def read_file(filename):
+    """
+    :returns: list of words in the file and removes any non-alpha-numeric item
+    #need to update with the - and -- that are in lines
+    """
+    #need to clean up input including punctuation
+    ln_lst = []
+    alpha_num = "abcdefghijklmnopqrstuvwxyz0123456789-."
```
nairraghav, line 14: I would recommend removing -. out of alphanum, or atleast .

students/charlie_england/session04/trigrams.py
```diff
@@ -0,0 +1,80 @@
...
+        else:
+            rtrn_lst.append(r_choice)
+        punc = 0
+        punc = punc_check(r_choice) #checks the puncation in the r_choice and changes puncuation state to 1 if there is puncuation
+    return " ".join(rtrn_lst)
+    
+
+if __name__ == "__main__":
+    words = read_file("sherlock.txt")
+    trigram = build_trigrams(words)
```
nairraghav, line 79: looks good!

nairraghav:
Looks good overall Charlie! I'm merging this for now since it contains the majority of what we're looking for. Please submit a new PR for the Mailroom 2 fix (try running with adding a new user with multiple donations to see the problem I mentioned). Also include the dict switch as well. Nice job!

PR #111, opened 2019-05-07
Assignment 4 submission

archwu:


students/ArchWu/Session4/dict_lab.py
```diff
@@ -0,0 +1,70 @@
...
+    Display the dictionary.
+    Add an entry for “fruit” with “Mango” and display the dictionary.
+    Display the dictionary values.
+    Display whether or not “cake” is a key in the dictionary (i.e. False) (now).
+    Display whether or not “Mango” is a value in the dictionary (i.e. True).
+
+    """
+
+    print(record1)
+    for v in record1:
```
nairraghav, line 17: You may want to consider using .pop or del here. Otherwise you have to loop through the entire dict to find what you are trying to delete.

Ex: dict.pop(key) 
del dict[key]

students/ArchWu/Session4/dict_lab.py
```diff
@@ -0,0 +1,70 @@
...
+        if 'cake' in v:
+            v.remove('cake')
+    print(record1)
+    record1['fruit'] = 'Mango'
+    print(record1)
+    print(record1.values())
+    exist_mango = False
+    for v in record1.values():
+        if 'cake' in v:
```
nairraghav, line 26: To check for a key in a dict, we don't need to go through the entire dict. You can just check: if 'cake' in dict

students/ArchWu/Session4/dict_lab.py
```diff
@@ -0,0 +1,70 @@
...
+        exist_mango = exist_mango or 'Mango' in v
+    print(exist_mango)
+
+def dict2():
+    """ dict tracking number of t's in keys """
+    print(record1)
+    record2 = {}
+    for k, v in record1.items():
+        count = 0
+        for letter in v:
```
nairraghav, line 37: instead of going through each character at once, you can use the .count() function to count how many t's there are in a string. You may want to lower the string so all t's are lower-case when searching

students/ArchWu/Session4/dict_lab.py
```diff
@@ -0,0 +1,70 @@
...
+
+def sets2():
+    s1 = {x for x in 'Python'}
+    s1.add('i')
+    s2 = frozenset({x for x in 'marathon'})
+    print(s1 | s2)
+    print(s1 & s2)
+
+
+if __name__ == '__main__':
```
nairraghav, line 65: Nice job!

students/ArchWu/Session4/file_lab.py
```diff
@@ -0,0 +1,57 @@
...
+                                lang_stats[lang] += 1
+                            else:
+                                lang_stats[lang] = 1
+                            language = "{}".format(lang)
+                            languages = languages + language
+                    result = name + languages
+                    print(result)
+        print(lang_stats)
+
+if __name__ == '__main__':
```
nairraghav, line 54: 👏 👏  nice!

students/ArchWu/Session4/mailroom.py
```diff
@@ -0,0 +1,116 @@
...
+    avg = round((total / num_gifts), 2)
+    return (name, total, num_gifts, avg)
+
+def thank_you():
+    """ Thank-you funcion that generates an email to thank new donations """
+    while True:
+        answer = input("Please type the Full Name, type quit to go back to previous menu> ")
+        if answer == 'list':
+            for donor in donors:
+                print(donor[0])
```
nairraghav, line 47: Be careful here, we are no longer using a tuple so we are not printing the 0th index of the string. We want the entire key 😄

students/ArchWu/Session4/mailroom.py
```diff
@@ -0,0 +1,116 @@
...
+                add_donation(name, num_donation)
+            break
+    greetings = 'Dear {},\n'.format(name)
+    body = '\nThank you for your generous gift ${} to us!\n'.format(num_donation)
+    ending = '\nSincerely,\n  ABC foundations'
+    print(greetings + body + ending)
+    return
+
+def thanks_all():
+    temp_path = tempfile.gettempdir()
```
nairraghav, line 79: You don't necessarily have to create this temp path, you can just get the current working directory. But nice!

nairraghav:
Great job Arch! Things look good overall. Please do look over some of the comments left about searching for keys in dicts. The benefit of using dicts is how fast you can find the key instead of iterating through each key one at a time. 

Also, try to resolve the small issue with printing only the first character of the name in mailroom when the user types "list".

PR #114, opened 2019-05-07
Finished three assignments, one more need to be done

momosally:
For the fileExercise one, I didn't quite get the requirements. So I add "None", if there is no nickname or language.

### DIFF
--- a/students/Sally_Shen/lesson4/dict_lab.py
+++ b/students/Sally_Shen/lesson4/dict_lab.py
@@ -1,0 +1,92 @@
+'''
+Dictionaries 1
+Create a dictionary containing “name”, “city”, and “cake” for “Chris” from “Seattle” who likes “Chocolate” (so the keys should be: “name”, etc, and values: “Chris”, etc.)
+Display the dictionary.
+Delete the entry for “cake”.
+Display the dictionary.
+Add an entry for “fruit” with “Mango” and display the dictionary.
+Display the dictionary keys.
+Display the dictionary values.
+Display whether or not “cake” is a key in the dictionary (i.e. False) (now).
+Display whether or not “Mango” is a value in the dictionary (i.e. True).
+'''
+
+dict1 = dict([('name', 'Chris'), ('city', 'Seattle'), ('cake', 'Chocolate')])
+print(dict1)
+
+
+del dict1['cake']
+print(dict1)
+
+dict1['fruit'] = 'Mango'
+print(dict1)
+
+
+print(dict1.keys())
+
+print(dict1.values())
+
+is_key = 'cake' in dict1.keys()
+print(is_key)
+
+is_value = 'Mango' in dict1.values()
+print(is_value)
+
+
+
+'''
+Dictionaries 2
+'''
+
+dict2 = {}
+
+for key in dict1.keys():
+    dict2[key] = dict1[key].lower().count('t')
+
+print(dict2.keys())
+print(dict2.values())
+
+
+'''
+Create sets s2, s3 and s4 that contain numbers from zero through twenty, divisible by 2, 3 and 4 (figure out a way to compute those – don’t just type them in).
+Display the sets.
+Display if s3 is a subset of s2 (False)
+and if s4 is a subset of s2 (True).
+'''
+s2 = set()
+s3 = set()
+s4 = set()
+
+for num in range(0,21):
+    if num % 2 ==0:
+        s2.add(num)
+    if num % 3 ==0:
+        s3.add(num)
+    if num % 4 ==0:
+        s4.add(num)
+print(s2)
+print(s3)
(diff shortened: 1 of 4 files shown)


students/Sally_Shen/lesson4/dict_lab.py
```diff
@@ -0,0 +1,92 @@
...
+Delete the entry for “cake”.
+Display the dictionary.
+Add an entry for “fruit” with “Mango” and display the dictionary.
+Display the dictionary keys.
+Display the dictionary values.
+Display whether or not “cake” is a key in the dictionary (i.e. False) (now).
+Display whether or not “Mango” is a value in the dictionary (i.e. True).
+'''
+
+dict1 = dict([('name', 'Chris'), ('city', 'Seattle'), ('cake', 'Chocolate')])
```
nairraghav, line 14: Another way to write this would be: dict1 = {"name":"Chris", "city":"Seattle", "cake":"Chocolate"}

students/Sally_Shen/lesson4/dict_lab.py
```diff
@@ -0,0 +1,92 @@
...
+
+dict1['fruit'] = 'Mango'
+print(dict1)
+
+
+print(dict1.keys())
+
+print(dict1.values())
+
+is_key = 'cake' in dict1.keys()
```
nairraghav, line 29: When looking for a key in a dict, you don't need to specify .keys(). 
Ex: 'cake' in dict1
This will return the same

students/Sally_Shen/lesson4/dict_lab.py
```diff
@@ -0,0 +1,92 @@
...
+add_letter = set('Python')
+add_letter.add('i')
+print(add_letter)
+
+
+frozen_set = frozenset('marathon')
+print(frozen_set)
+
+print('Union of the two sets is', add_letter.union(frozen_set))
+print('Intersection of the two set is', add_letter.intersection(frozen_set))
```
nairraghav, line 91: Nice job!

students/Sally_Shen/lesson4/fileExercise.py
```diff
@@ -0,0 +1,99 @@
...
+            values = line.strip().split(":")
+            if not values[1]:
+                # no potential language list
+                continue
+            else:
+                potentialLanguages = values[1]
+                for potentialLanguage in potentialLanguages.split(","):
+                    potentialLanguage = potentialLanguage.strip()
+                    #print("potentialLanguage:" + potentialLanguage)
+                    if potentialLanguage in ExistingLanguages:
```
nairraghav, line 53: The way to tell if the string is potential language is that the languages are all in lowercase while the nicknames are uppercase. So what you want to do is validate if potentialLanguage is capitalized (check if the first letter is upper case). 

Otherwise,  if we have a language not specified above, it will skip it.

students/Sally_Shen/lesson4/mailroom2.py
```diff
@@ -0,0 +1,123 @@
...
+    print(output)
+
+def lineItem(name, total, numGifts, avg):
+    output = "{:<20}${:<20.2f}{:<15}${:<20.2f}\n".format(name, total, numGifts, avg)
+    return output
+
+def gen_thank_you_letters(a_dict):
+    dir = "thankyouLetters"
+    try:
+        os.mkdir(dir)
```
nairraghav, line 70: Once you make this directory, you will want to enter it. Otherwise all the files created are in the same directory (outside of thankyouLetters)

students/Sally_Shen/lesson4/mailroom2.py
```diff
@@ -0,0 +1,123 @@
...
+        answer = input ("""What would you like to do?
+    Pick one:
+    1. Send a Thank You to a single donor
+    2. Create a Report
+    3. Send Thankyou letters to all donors
+    4. Quit
+    >>>""")
+        print("You picked:", answer)
+        answer = answer.strip()
+        if answer == "1":
```
nairraghav, line 106: Try to use a dict switch here!

students/Sally_Shen/lesson4/mailroom2.py
```diff
@@ -0,0 +1,123 @@
...
+
+    if response not in donors.keys():
+        donors[response] = []
+
+    newDonation = input("how much to donate?")
+    donors[response].append(float(newDonation))
+    email = '''
+    Dear {0},
+    Thank you for donating {1}!\n'''.format(response, newDonation)
+    return email
```
nairraghav, line 32: After returning email here, you are not printing it. You should be printing a thank you note after each individual donation.

nairraghav:
Overall, everything looks good! Please look at my comments for the mailroom2, there are some small changes you will want to make.

PR #118, opened 2019-05-08
I found the error, it should work

gjkim44:


students/GKim/lesson04/mailroom2.py
```diff
@@ -114,7 +114,7 @@ def send_thank_you():
             
             if in_main_donors == False:
                 donation_amount = float(input("\nPlease enter {}'s donated amount: ".format(thanks_answer)))
-                add_new_donors = {"name": thanks_answer, "donation": [donation_amount]}
+                add_new_donors = {"name": thanks_answer, "donations": [donation_amount]}
```
nairraghav, line 117: 👌

nairraghav:
Nice job! Just a small note: your "Send to All" copies the same template so there's a wording issue where it says: "Thank you for your donation _totaling_ ..." but you are only printing out the last donation amount. Not a breaking issue or anything, you don't have to fix it but just lettin' ya know 😀

PR #119, opened 2019-05-10
Completed file exercise 

jchando1:


students/JasneetChandok/Lesson04/fileexcercise.py
```diff
@@ -0,0 +1,47 @@
...
+folder_path = os.getcwd()
+list_of_file = glob.glob(os.path.join(folder_path, '*.py'))
+
+for file_name in list_of_file:
+    print(file_name)
+
+
+# Part 2
+path = pathlib.Path.cwd()
+sourcefile_path = 'C:\\Users\\jchando1\\OneDrive - T-Mobile USA\\Programming in Python\\SP_2019_210A_classroom\\students\\JasneetChandok\\Readme.txt'
```
nairraghav, line 21: Be careful here, you don't want to hardcode your path. You want to use the os.path to find the path of your file (it is usually easier to put the file into your current working directory).

students/JasneetChandok/Lesson04/fileexcercise.py
```diff
@@ -0,0 +1,47 @@
...
+path2 = path / "students.txt" # directing the path to your student directory
+with open(path2) as stdfile:
+    stdfile.readline()
+    for rows in stdfile:
+        langs = rows.split(':')[1]  
+        langs = langs.replace(',', ' ')
+        langs = langs.replace('and', ' ')
+        langs = langs.split()
+        for lang in langs:
+            if lang:
```
nairraghav, line 40: You are not using any logic here to filter out the nickname of the student. This will still grab the nickname if it exists. Notice that the nicknames are capitalized while the languages are not. Think about how you can ignore the nicknames in the list. 

Since you're pretty much there, I will give you credit but try to take a look at how you could resolve this issue!

students/JasneetChandok/Lesson04/mailroom2.py
```diff
@@ -0,0 +1,172 @@
+#!/usr/bin/env python3.6
+
+'''Lesson 3 - Mailroom Part 1'''
```
nairraghav, line 3: Part 2 😄

students/JasneetChandok/Lesson04/mailroom2.py
```diff
@@ -0,0 +1,172 @@
+#!/usr/bin/env python3.6
+
+'''Lesson 3 - Mailroom Part 1'''
+
+
+# Importing Systems Module
+import sys
+
+# Starter list of existing donors
+donors = {'jasneet paramjit': ("jasneet paramjit", [50, 200, 500]),
```
nairraghav, line 10: You actually don't need to have the tuple as the value anymore, it just adds more clutter. You can just have the array as the value:

{'jasneet paramjit': [50, 200, 500], 'simran kaur': [100, 200, 500]}

This will make it slightly easier when grabbing/updating the values

students/JasneetChandok/Lesson04/mailroom2.py
```diff
@@ -0,0 +1,172 @@
...
+
+# Function for sending a thanking you note
+def thankyou_email():
+    donor_category = input("""Specify if new or pick existing donor from the list
+        New Donor: Enter 'new'
+        Existing Donor: Enter 'list'
+        >>>""")
+    donor_category = donor_category.strip().lower()
+    print("\nYou selected:", donor_category, "\n")
+    if donor_category == "new":
```
nairraghav, line 79: You actually don't want to *force* them into using an existing user if they type list. An easy way to remedy this is to take the logic of the "show list" out of the existing_donor. Show the list, then do the pick depending on the donor name being in the dict.

students/JasneetChandok/Lesson04/mailroom2.py
```diff
@@ -0,0 +1,172 @@
...
+
+    selection_dict = {"1": thankyou_email,
+                      "2": generate_report,
+                      "3": save_letters_to_disk,
+                      "4": quit_program}
+
+    run_menu(prompt, selection_dict)
+
+
+def run_menu(prompt, selection_dict):
```
nairraghav: I'm not able to get this working when I try to print the report. Try running it again and see if you're able to get this behavior to work.

nairraghav:
Good job Jasneet! I was not able to get the mailroom part 2 fully functional, please look at the comment I've left for that method. You're almost there! Since your changes are a bit minimal, please address the sort/print issue and I will merge this in.

students/JasneetChandok/Lesson04/mailroom2.py
```diff
@@ -0,0 +1,172 @@
...
+    report_rows.sort(key=sort_key)
+    report = []
+    report.append("{:25s} | {:11s} | {:9s} | {:12s}".format("Donor Name",
+                                                            "Total Given",
+                                                            "Num Gifts",
+                                                            "Average Gift"))
+    report.append("-" * 66)
+    for row in report_rows:
+        report.append("{:25s}   ${:10.2f}   {:9d}   ${:11.2f}".format(*row))
+    return "\n".join(report)
```
nairraghav, line 111: I believe your issue is that you are not printing anything :) 
Instead of returning, do a print here.

students/JasneetChandok/Lesson04/mailroom2.py
```diff
@@ -0,0 +1,172 @@
...
+def generate_report():
+    report_rows = []
+    for (name, gifts) in donors.values():
+        total_gifts = sum(gifts)
+        num_gifts = len(gifts)
+        avg_gift = total_gifts / num_gifts
+        report_rows.append((name, total_gifts, num_gifts, avg_gift))
+
+    # sort the report data
+    report_rows.sort(key=sort_key)
```
nairraghav, line 102: Sort defaults to ascending so your lowest value is first, add a reverse=true to flip it!

nairraghav:
Hi @jchando1, I'm going to merge this in for now. Please review the comments above and address them in a future pull request.

PR #121, opened 2019-05-11
adding mailroom1.py

BBrumleve:
working on adding dictionaries now.
thanks,
Brian


students/BrianB/lesson03/mailroom1.py
```diff
@@ -1,121 +1,109 @@
...
-                                                            employer
-                                '''.format(donor[0], donor[1][-1])))
-                break  # this does not break out of [thank =_you()]
+                print(donor[0], "was found in our database!")
+                new_amt = input("Please enter a new donation amount:")
+                donor[1].append(int(new_amt))
+                print(donor_data)
+                start_over = input("If you would like to try another "
+                                   "donor enter 'yes' or 'no'.")
+                if start_over.lower() == "yes":
```
nairraghav, line 53: Be careful here because the break here kicks you out of the forloop but not the while loop.
If you enter a new user and type in "no" for the start_over, it still prompts you for another. 

You will want to put a break in the for loop where you find the user and the if start_over logic just outside the for loop

nairraghav:
This looks pretty good Brian, please address the issue I've noted above! I think once that is resolved, this should be complete.

PR #122, opened 2019-05-11
Adding session05 files

ervbrian:
Added exceptions lab and mailroom.py with exception handling / comprehensions.

### DIFF
--- a/students/BrianE/session05/MailRoom.py
+++ b/students/BrianE/session05/MailRoom.py
@@ -1,0 +1,126 @@
+#!/usr/bin/env python3
+
+import sys
+import tempfile
+
+'''
+Manage donation information and generate thank-you emails for donors
+'''
+
+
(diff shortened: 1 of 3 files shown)


students/BrianE/session05/MailRoom.py
```diff
@@ -0,0 +1,126 @@
...
+def generate_all_thankyou_emails(donors):
+    """
+    Generate thank-you emails to all donors in database
+    :param donors: dictionary of donors
+    :return: write email text to file for each donor
+    """
+
+    for donor in donors:
+        total_donated = round(sum(donors[donor]), 2)
+        most_recent_donation = donors[donor][0]
```
nairraghav, line 50: Your most recent donation would actually be the last item in the list!

nairraghav:
Nice job Brian!

PR #129, opened 2019-05-13
Add two exception excerises

momosally:
Still working on the other two assignments.

### DIFF
--- a/students/Sally_Shen/lesson5/MailroonPart3.py
+++ b/students/Sally_Shen/lesson5/MailroonPart3.py
@@ -1,0 +1,142 @@
+
+'''choms'''
+
+import sys
+import os
+
+print("in mailroom")
+'''Writing first name is easier for testing.'''
+donors ={"Fred Gates": [100.90, 200, 50], "Jeff Bezos": [100, 50, 600], "Mark Zuckerberg": [200, 50, 500],
+          "Paul Allen": [300, 40, 200]}
+
+
+# donors = {{"first_name": "Fred", "last_name": "Gates"}: [[100.90, 200, 50], 265.98]}
+
+
+def thank_you():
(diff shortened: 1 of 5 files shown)


students/Sally_Shen/lesson5/MailroonPart3.py
```diff
@@ -0,0 +1,142 @@
...
+def lineItem(name, total, numGifts, avg):
+    output = "{:<20}${:<20.2f}{:<15}${:<20.2f}\n".format(name, total, numGifts, avg)
+    return output
+
+
+def gen_thank_you_letters(a_dict):
+    dir = "thankyouLetters"
+    try:
+        os.mkdir(dir)
+    except FileExistsError:
```
nairraghav, line 91: nice!

students/Sally_Shen/lesson5/MailroonPart3.py
```diff
@@ -0,0 +1,142 @@
...
+        pass
+    for name, amount in donors.items():
+        firstLastName = name.split()
+        fileName = firstLastName[0] + "_" + firstLastName[1] + ".txt"
+        filePath = dir + "\\" + fileName
```
nairraghav, line 96: This ends up prepending dir + \\ into the filename. What you want to do is change the working directory to enter the filePath and then create the files.
For example:
os.chdir(path)

nairraghav:
Great job Sally! I've added a comment on how to add your created files for Mailroom into the directory that you created. You can try to implement that as a part of your next mailroom

students/Sally_Shen/lesson5/MailroonPart3.py
momosally: Thank you, fixed in lesson 6 PR

PR #130, opened 2019-05-13
init commit of Assignment 5

archwu:


students/ArchWu/Session5/except_exercise.py
```diff
@@ -0,0 +1,58 @@
...
+langs = ['java', 'c', 'python']
+
+try:
+    more_joke = more_fun(langs[0])
+except IndexError:
+    pass
+try:
+    more_jokez = more_fun(langs[1])
+except:
+    pass
```
nairraghav, line 58: You're just missing the last_fun call. but there is no additional logic here so I'll give you credit for the assignment!

students/ArchWu/Session5/mailroom.py
```diff
@@ -0,0 +1,114 @@
...
+def thank_you():
+    """ Thank-you funcion that generates an email to thank new donations """
+    while True:
+        answer = input("Please type the Full Name, type quit to go back to previous menu> ")
+        if answer == 'list':
+            for donor in donors:
+                print(donor)
+        elif answer == 'quit':
+            return
+        elif answer in [donor[0] for donor in donors]:
```
nairraghav, line 48: Be careful, you don't want to do this. donor[0] for donor in donors is the first character of each donor. I don't believe this is your intended action here.

nairraghav: You do not need a comprehension here. You only want to check if answer is in donors

nairraghav:
I'm going to merge this for now, please take a look at the comment in mailroom and submit another pull request with that fix.

PR #131, opened 2019-05-14
PR for Lesson05

miotke:
Added a few try/excepts into mailroom. 
Also cleaned up and finished up some past labs and excerises.


students/AndrewMiotke/session03/list_lab.py
```diff
@@ -1,21 +1,49 @@
...
+fruit_list.pop()
+print(fruit_list)
+
+delete_fruit = input("What fruit would you like to delete? ")
+fruit_list.remove(delete_fruit)
+print(fruit_list)
+
+# Series Three
+while True:
+    for favorite_fruit in fruit_list:
```
nairraghav, line 38: Be careful here, if you remove the fruit while iterating, you'll actually skip head. Try running this and see the result when you remove a fruit (you'll end up skipping the next one in line)

nairraghav: To resolve this, you'll want to iterate through the list in a way which lets you keep track of the index

students/AndrewMiotke/session05/except_exercise.py
```diff
@@ -0,0 +1,53 @@
...
+# If there are no exceptions, call the more_fun function with the last
+# language in the list
+
+# Finally, while still in the try/except block and regardless of whether
+# there were any exceptions, call the function last_fun with no
+# parameters. (pun intended)
+
+langs = ['java', 'c', 'python']
+
+more_joke = more_fun(langs[0])
```
nairraghav, line 53: Don't forget to put this into a try/except block!

nairraghav: I believe you are missing some other fields as well, please revisit this exercise and add the remaining bits

students/AndrewMiotke/session05/mailroom.py
```diff
@@ -0,0 +1,172 @@
+#!/usr/bin/env python3
+import os
+import sys
+import time
+import shutil
+from textwrap import dedent # I took this from the solutions code because I thought it was great
+
+# The names come from dogs that are in my office
```
nairraghav, line 8: 🐶

students/AndrewMiotke/session05/mailroom.py
```diff
@@ -0,0 +1,172 @@
...
+            main_menu()
+        else:
+            break
+
+    while True:
+        try:
+            add_amount = input("Enter an amount to donate: ")
+            donated_amount = float(add_amount)
+            break
+        except ValueError:
```
nairraghav, line 74: Nice handling!

students/AndrewMiotke/session05/mailroom.py
```diff
@@ -0,0 +1,172 @@
+#!/usr/bin/env python3
+import os
+import sys
+import time
+import shutil
+from textwrap import dedent # I took this from the solutions code because I thought it was great
+
+# The names come from dogs that are in my office
+donors_list = [("Rufio", [897, 200, 200]),
```
nairraghav, line 9: By now, you should try to convert this into a dict. It will make it much easier to go through and find if a user exists instead of iterating through the list every time you want to do something.

students/AndrewMiotke/session05/mailroom.py
```diff
@@ -0,0 +1,172 @@
...
+
+            answer_dict = {
+                "1": send_thank_you_letter,
+                "2": report,
+                "3": send_letter_to_all_donors,
+                "4": quit,
+            }
+
+            answer_dict.get(answer)()
+        except TypeError:
```
nairraghav, line 163: 👏 👏

nairraghav:
Good job with the exceptions, please do try to revisit the mailroom using a dict instead of a list. This will make your life a bit easier otherwise you have to iterate through the list every time you want to do an action with a specific user. Also take a look at list_lab again, I believe you have everything except the remove for that exercise. Finally, review the except_exercise, it seems you are missing some parts of the assignment.

There is a lot of work here so I will merge this in for now, please address the changes in a new PR.

PR #133, opened 2019-05-14
HW Done

FredBallyns:
Ignore Dict_Lab and Mailroom2. Not sure why they still showing.

I'm curious if there is a more pythonic way of doing a long multi "or" statement. Would you suggest using a dictionary?

Also couldn't get "code 'and not' code" to work so instead did "'not' code 'and' code" . Any advise how should have typed?


students/FredB/Fred_Presentation.py
```diff
@@ -0,0 +1,104 @@
...
+        remain_text = len(remain_text)
+    remain_text = int(remain_text)
+    len_str=len(input_text)
+    for i in range(len_str):
+        temp=input_text[:(len_str-i-1+remain_text)]
+        time.sleep(.05)
+        clear_screen()
+        print(temp, end="")
+
+def hype_intro():
```
nairraghav, line 44: 🎉🎉🎉🎉

students/FredB/session05/except_exercise.py
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/python
```
nairraghav, line 1: Nice job!

students/FredB/session05/mailroom3.py
```diff
@@ -0,0 +1,152 @@
...
+#Session04
+
+import sys
+import os
+
+donors = {"Fred Flintstone": [100, 200, 50],
+        "James Bond": [0,0,7],
+        "Lex Luthor": [5000,],
+        "Harambe": [3, 2, 1],
+        "Herman Cain": [9, 9, 9]
```
nairraghav, line 13: You can just put the } at the end of the line here

students/FredB/session05/mailroom3.py
```diff
@@ -0,0 +1,152 @@
...
+        #    donors[donor] = [donation]
+        #print("donors",'\n','------------\n',donors)
+        donors.setdefault(donor, []).append(donation)
+        print (f"\n Thank you {donor}, for your generous donation of ${donation} from a charity \n")
+
+
+def check_donor(donor):
+        try:
+            len(donors[donor])
+        except KeyError:
```
nairraghav, line 58: Interesting way to handle this!

students/FredB/session05/mailroom3.py
```diff
@@ -0,0 +1,152 @@
+#!/usr/bin/env python3.7
```
nairraghav, line 1: Looks good!

nairraghav:
I'm curious if there is a more pythonic way of doing a long multi "or" statement. Would you suggest using a dictionary?
> I think it would depend on the situation. Can you give me an example of a multi "or" statement that you were running into?

Also couldn't get "code 'and not' code" to work so instead did "'not' code 'and' code" . Any advise how should have typed?
> I'm not sure what you're referring to here. Can you give me an example?

PR #136, opened 2019-05-17
Reworked mailroom assignment 2

jchando1:


students/JasneetChandok/Lesson04/trigrams.py
```diff
@@ -0,0 +1,63 @@
...
+punctuations = "!" + "-" + "," + "." + "(" + ")" + '"' + '?' + ';' + ":" + '\n'
+replacements = " " * 11
+
+
+def read_file(filename):
+    words = []
+    with open(filename, "r") as text:
+        for line in text.readlines():
+            raw_line = line
+            clean_string = raw_line.translate(raw_line.maketrans(punctuations,replacements))
```
nairraghav, line 20: Nice clean up here!

Instead of creating your own punctuations string above, you can use a predefined string.punctuation:


import string
string.punctuation #this is a string that contains all that you would need

students/JasneetChandok/Lesson04/trigrams.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+def main():
+    in_file = read_file("sherlock_small.txt")
+    trigrams = build_trigrams(in_file)
+    final_text = make_text(trigrams)
+    print(final_text)
+
+
+if __name__ == "__main__":
```
nairraghav, line 62: well done!

students/JasneetChandok/Lesson04/mailroom2_v2.py
```diff
@@ -0,0 +1,180 @@
...
+# List of existing donor
+def existingdonors_list():
+    donor_list = "\n ".join(donor for donor in donors)
+    print("\nList of Existing Donors:", "\n", donor_list.title(), "\n")
+
+
+# Add donation to an existing donor
+def add_donation():
+        donor_selection = input("Enter 'Full Name' of the donor >")
+        donor_donation = input("\nEnter the 'Amount' donated >")
```
nairraghav, line 26: you can actually convert this to an int here itself:
donor_donation = int(input("Enter Donation Amount"))

students/JasneetChandok/Lesson04/mailroom2_v2.py
```diff
@@ -0,0 +1,180 @@
...
+        donor = donor_selection.strip().lower()
+        if donor in donors:
+            donor_amount = donors.get(donor)
```
nairraghav, line 29: To append here, you don't need to create a new variable, you can append directly, like so:
donors[donor].append(donation_amount)

students/JasneetChandok/Lesson04/mailroom2_v2.py
```diff
@@ -0,0 +1,180 @@
...
+        else:
+            print('\nDonor mentioned does not exist in the list!!')
+
+
+# Create a record for a new donor
+def new_donor():
+        donor_name = input("Enter 'Full Name' of the donor >")
+        donor_name = donor_name.strip().lower()
+        new_donation = int(input("\nEnter the 'Amount' donated >"))
+        if donor_name not in donors:
```
nairraghav, line 50: Since you are calling this function "new_donor", you should already know that the donor_name doesn't exist in donors. This doesn't need to be called here

students/JasneetChandok/Lesson04/mailroom2_v2.py
```diff
@@ -0,0 +1,180 @@
...
+            donors[donor_name] = [new_donation]
+            print(f'\n Donation amount of "{new_donation}" has been recorded for {donor_name.title()}!!\n')
```
nairraghav, line 52: Since this message seems to be a duplicate of the message printed in the event of an existing user, you may want to consider having  a function handle the printing so you can call it for both functions, instead of having the same code copied.

students/JasneetChandok/Lesson04/mailroom2_v2.py
```diff
@@ -0,0 +1,180 @@
...
+    report.append("-" * 70)
+
+# print the report
+    for row in report_rows:
+        report.append("{:20s}   {:10.2f}$   {:9d}   {:13.2f}$".format(*row))
+    print("\n".join(report))
+
+
+def gen_letter(donor):
+    return ('''Dear {:s},
```
nairraghav, line 118: It looks like you already have a function here, you should use this above :)

students/JasneetChandok/Lesson04/mailroom2_v2.py
```diff
@@ -0,0 +1,180 @@
...
+    selection_dict = {"1": thankyou_email,
+                      "2": generate_report,
+                      "3": save_letters_to_disk,
+                      "4": quit_program}
+
+    run_menu(prompt, selection_dict)
+
+
+# Function to enabling the main menu option
+def run_menu(prompt, selection_dict):
```
nairraghav, line 163: Nice job making this method to run in multiple places!

students/JasneetChandok/Lesson04/mailroom2_v2.py
```diff
@@ -0,0 +1,180 @@
...
+                break
+
+
+# Runs the file as a main module when opened by the user
+if __name__ == "__main__":
+    print("\n*****Welcome to the Mailroom Program!!*****\n")
+
+
+# Call funcion for the main menu when run manually
+# main_menu()
```
nairraghav, line 180: Don't forget to uncomment this to actually run!

nairraghav: Overall this looks great, nice job

students/JasneetChandok/Lesson05/except_exercise_updated.py
```diff
@@ -0,0 +1,57 @@
...
+
+langs = ['java', 'c', 'python']
+try:
+    more_joke = more_fun(langs[0])
+except IndexError:
+    more_joke = more_fun(langs[1])
+else:
+    more_joke = more_fun(langs[2])
+finally:
+    last_fun()
```
nairraghav, line 57: Great job!

students/JasneetChandok/Lesson05/mailroom3.py
```diff
@@ -0,0 +1,186 @@
...
+        else:
+            if action():
+                break
+
+
+# Runs the file as a main module
+if __name__ == "__main__":
+    print("\n*****Welcome to the Mailroom Program!!*****\n")
+
+main_menu()
```
nairraghav, line 186: This should go under your "welcome to mailroom"

students/JasneetChandok/Lesson05/mailroom3.py
```diff
@@ -0,0 +1,186 @@
...
+
+# Add donation to an existing donor
+def add_donation():
+    try:
+        donor_selection = input("Enter 'Full Name' of the donor >")
+        donor_donation = input("\nEnter the 'Amount' donated >")
+        donor = donor_selection.strip().lower()
+        if donor in donors:
+            donor_amount = donors.get(donor)
+            donor_amount.append(int(donor_donation))
```
nairraghav, line 34: You will want to be careful here. You can run into an exception if the donor_donation is a string!

students/JasneetChandok/Lesson05/mailroom3.py
```diff
@@ -0,0 +1,186 @@
...
+        else:
+            print('\nDonor name entered does not exist in the list!!')
+    except (EOFError, KeyboardInterrupt):
+        print("""\nA user induced error occured. Please check keys pressed!""")
+
+# Create a record for a new donor
+def new_donor():
+    try:
+        donor_name = input("Enter 'Full Name' of the donor >")
+        new_donation = int(input("\nEnter the 'Amount' donated >"))
```
nairraghav, line 54: Same goes for here, you can run into  an error if you input a string instead of int

nairraghav:
Nice job Jasneet! I am giving you credit for everything here except the Mailroom Part 3. Please look to adding some exceptions for the lines where you convert the input into an int. If you pass a string, it will throw an exception. I believe you have the idea down but it would be good if you can add that!

PR #137, opened 2019-05-18
Trigrams.py and Mailroom_part4 with their test files

kanmenlaohan:


students/Lincoln_Zhang/Session4/test_trigrams.py
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env python3
+import trigrams,os
+
```
PythonCHB, line 3: great idea to test this!

students/Lincoln_Zhang/Session4/test_trigrams.py
```diff
@@ -0,0 +1,14 @@
...
+global_result = trigrams.read_file("sherlock_small.txt")
+
+def test_read_file():
```
PythonCHB, line 6: the goal is to have unit test be stand-alone -- so better to have the test(s) that test , e.g. file reading, actually read the file as part of the test, rather than as a global.

students/Lincoln_Zhang/Session4/test_trigrams.py
```diff
@@ -0,0 +1,14 @@
...
+	assert global_result[0] == "One"
+	assert global_result[1] == "night"
+
+def test_build_trigram():
+	local_result = trigrams.build_trigram(global_result)
```
PythonCHB, line 11: here is where you can really isolate the tests. 

In this case, you want to test build_tirgram with a short, simple text, rather than ven the "small" real text, Perhaps:

"I wish I may I wish I might"

or a made up text that will test what you want. -- maybe one, two and three "followers"

students/Lincoln_Zhang/Session4/test_trigrams.py
```diff
@@ -0,0 +1,14 @@
...
+	assert type(local_result) == dict
```
PythonCHB, line 12: you generally don't want to test that you get the right type -- do you care? what you want is an object with the right interface -- that is, use the result, and see if it works. (like you have in the following lines)

students/Lincoln_Zhang/Session4/test_trigrams.py
```diff
@@ -0,0 +1,14 @@
...
+	assert local_result[('with', 'a')] == ['keen']
+	assert local_result[('One', 'night')] == ['it']
```
PythonCHB, line 14: you should hvae a test that checks that it properly reads more than one following word.

students/Lincoln_Zhang/Session4/trigrams.py
```diff
@@ -5,23 +5,35 @@
...
-words = "I wish I may I wish I might".split()
+# words = "I wish I may I wish I might".split()
 
-for w in zip(words,islice(words,1,None),islice(words,2,None))
-    print(w)
+# for w in zip(words,islice(words,1,None),islice(words,2,None)):
+#     print(w)
 
 
 def read_file(fname):
```
PythonCHB, line 18: you tests for this function could make sure punctuation  was handles properly, etc.

students/Lincoln_Zhang/Session4/trigrams.py
```diff
@@ -45,18 +57,27 @@ def build_trigram(words):
     return trigrams
 
 def text_gen():
```
PythonCHB, line 59: this is really hard to test, as it relies on random numbers -- so the results aren't consistent :-)

but are there some things you could test about it?

students/Lincoln_Zhang/Session6/test_mailroom_part4.py
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env python3
+
+import mailroom_part4, os
+import unittest.mock
+
+def test_gen_letter():
+    result = mailroom_part4.gen_letter("Bob", 11.2)
+    assert result.startswith("\nDear Bob")
+    assert result.find("11.2")
```
PythonCHB, line 9: careful here! .find() returns the index of what it's found. and as zero is a valid index, it used -1 to mean "not found" -- but -1 is "truthy" -- so this assertion will always pass!

`assert "11.2" in result`

is a better check -- better still, something like:

`assert "$11.20" in result`

so you can see that it was formatted correctly.

students/Lincoln_Zhang/Session6/test_mailroom_part4.py
```diff
@@ -0,0 +1,55 @@
...
+
+#     file_path = os.path.join(os.getcwd(),"Bob.txt")
+#     expanded_file_path = os.path.expanduser(file_path)
+#     assert os.path.isfile(file_path)
+#     assert os.path.isfile(expanded_file_path)
+#     assert result["Bob"] == [11.2]
+
+def test_gen_stats():
+    result = mailroom_part4.gen_stats([1,2,3])
+    assert result[:2] == (6,3)
```
PythonCHB, line 51: nice!

students/Lincoln_Zhang/Session6/test_mailroom_part4.py
```diff
@@ -0,0 +1,55 @@
...
+
+def test_gen_report():
+    mailroom_part4.gen_report({"Bob":[11.2]})
+    assert row == "Donor Name            | Total Given |  Num Gifts  | Average Gift"
```
PythonCHB, line 55: row hasn't been defined has it?

always run your code before committing it!

PythonCHB (commented):
nice work -- getting the hand of testing.

PythonCHB:
I'm merging but not giveing credit for Mailroom 4 yet -- looks like you aren't done.

Reminder: don't do a PR until it's ready for review (either because you are done, or you have a question -- if you have  questions, then be sure to put them in the PR.

PR #138, opened 2019-05-19
Session 06 

miotke:
* Wrote some basic tests for mailroom
* Completed some older assignments


students/AndrewMiotke/session03/list_lab.py
```diff
@@ -34,14 +34,15 @@
...
 # Series Three
-while True:
-    for favorite_fruit in fruit_list:
-        users_favorite_fruit = input(f"do you like {favorite_fruit.lower()}, yes or no? ")
-        if users_favorite_fruit == "no":
-            fruit_list.remove(favorite_fruit)
-            print(fruit_list)
-        elif users_favorite_fruit != "yes" or "no":
-            print(fruit_list)
+for favorite_fruit in fruit_list:
```
nairraghav, line 37: You will want to be careful iterating here. If you run this and say no, you will delete the fruit the it will "skip" the next fruit in line. Try this out and see what happens

students/AndrewMiotke/session05/except_exercise.py
```diff
@@ -49,5 +44,11 @@
...
-
-more_joke = more_fun(langs[0])
+try:
+    more_joke = more_fun(langs[0])
+except IndexError:
+    more_joke = more_fun(langs[1])
+else:
+    more_joke = more_fun(langs[-1])
+finally:
+    last_fun
```
nairraghav, line 54: Nice job!

students/AndrewMiotke/session06/mailroom.py
```diff
@@ -0,0 +1,175 @@
+#!/usr/bin/env python3
+import os
+import sys
+import time
+import shutil
+from textwrap import dedent
+
+
+donors_list = {"Rufio": ("Rufio", [897, 200 , 200]),
```
nairraghav, line 9: To make this a bit simpler, you can remove the tuple and leave just the list of donations as the values:

{"Rufio": [897, 200 , 200], ... }

students/AndrewMiotke/session06/walnut_party/walnut_party.py
```diff
@@ -0,0 +1,26 @@
...
+40 and 60, inclusive. Unless it is the weekend, in which case there
+is no upper bound on the number of walnuts.
+
+Return True if the party with the given values is successful,
+or False otherwise.
+"""
+
+
+def walnut_party(walnuts, is_weekend):
+    if walnuts <= 30 and is_weekend == False:
```
nairraghav, line 15: This can be done much neater if you only apply the logic to return true and return false all other times. Ex:
if weekend and walnuts >= 40:
return true #handle weekend true case
if not weekend and walnuts >= 40 and walnuts <= 60:
return true #handles not weekend true case

return false #this will return false if our two options are not true

nairraghav:
Nice job with the testing Andrew! Check out my comment for list lab + walnuts exercise. Walnuts isn't graded but just a note on how to handle that logic a bit neater. When dealing with returning true/false, usually it's simpler to handle either true/false and then always return the opposite when those cases are not fulfilled. Let me know if you would like me to further clarify on that!

PR #140, opened 2019-05-20
Week 6 Homework

roviram:
Mailroom Part 4
Tests for Mailroom Part 4
Args/Kwargs Lab
Intro to Unit Testing Labs (Walnut Party + 2 additional from Coding Bat)

Thanks!

### DIFF
--- a/students/roviram/lesson06/caught_speeding.py
+++ b/students/roviram/lesson06/caught_speeding.py
@@ -1,0 +1,44 @@
+#---------------------------------------------------#
+# Dev: Miguel Rovira-Gonzalez
+# Week 6 Homework
+# Script Name: Caught Speeding
+# Script Creation Date: Saturday 5/18/2019
+#---------------------------------------------------#
+"""
+You are driving a little too fast, and a police officer stops you.
+Write code to compute the result, encoded as an int value: 0=no ticket, 1=small ticket, 2=big ticket.
+    If speed is 60 or less, the result is 0.
+    If speed is between 61 and 80 inclusive, the result is 1.
+    If speed is 81 or more, the result is 2.
+    Unless it is your birthday -- on that day, your speed can be 5 higher in all cases.
+
+caught_speeding(60, False) → 0
+caught_speeding(65, False) → 1
+caught_speeding(65, True) → 0
+"""
+
+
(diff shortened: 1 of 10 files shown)


students/roviram/lesson06/caught_speeding.py
```diff
@@ -0,0 +1,44 @@
...
+    If speed is 81 or more, the result is 2.
+    Unless it is your birthday -- on that day, your speed can be 5 higher in all cases.
+
+caught_speeding(60, False) → 0
+caught_speeding(65, False) → 1
+caught_speeding(65, True) → 0
+"""
+
+
+def caught_speeding(speed, is_birthday):
```
nairraghav: You can apply this logic a bit neater. One way to think about this is to try and only handle the True or False cases (just one of them). Once you apply those and if you still haven't returned out as true, then return False.

Ex:

if blah and blah: #truth case
return true

if blah and blah2: #second truth case
return true

if blah and blah3: #last truth case
return true

return false #if we haven't returned true by now, we know that we MUST be false

students/roviram/lesson06/squirrel_play.py
```diff
@@ -0,0 +1,19 @@
...
+    Unless it is summer, then the upper limit is 100 instead of 90.
+    Given an int temperature and a boolean is_summer,
+    return True if the squirrels play and False otherwise.
+squirrel_play(70, False) → True
+squirrel_play(95, False) → False
+squirrel_play(95, True) → True
+"""
+
+def squirrel_play(temp, is_summer):
+  return (60 <= temp <= 100 and is_summer) or (60 <= temp <= 90 and not is_summer)
```
nairraghav, line 19: neat!

students/roviram/lesson06/test_mailroom_part4.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def test_send_thank():
+    """This is testing the donor exists in my donors diotionary based off my send thank you method"""
+    result = mailroom.donors
+    assert "Miguel Rovira" in result
+    assert "Pepper" in result
+    assert "Cristina Rovira" in result
+    assert "Coffee" not in result
+
```
nairraghav, line 36: Nice job!

PR #143, opened 2019-05-20
Mailroom Part4 and test_mailroom.py

Charlie-England:
refactored mailroom to be testable as well as added tests in the test_mailroom.py file


nairraghav:
Nice job Charlie!

PR #147, opened 2019-05-23
Help Needed with HTML. Please review trigrams and file_exercises for score

BBrumleve:
Please focus on trigrams, file exercises and need help with HTML.

HTML errors:
==================================================================== test session starts =====================================================================
platform darwin -- Python 3.7.3, pytest-4.5.0, py-1.8.0, pluggy-0.11.0
rootdir: /Users/brianbrumleve/SP_2019_210A_classroom/students/BrianB/lesson07
collected 31 items                                                                                                                                           

test_html_render.py ...........F........F..........                                                                                                    [100%]

========================================================================== FAILURES ==========================================================================
______________________________________________________________________ test_attributes _______________________________________________________________________

    def test_attributes():
        e = P("A paragraph of text", style="text-align: center", id="intro")
    
        file_contents = render_result(e).strip()
        print(file_contents)  # so we can see it if the test fails
    
        # note: The previous tests should make sure that the tags are getting
        #       properly rendered, so we don't need to test that here.
        #       so using only a "P" tag is fine
        assert "A paragraph of text" in file_contents
        # but make sure the embedded element's tags get rendered!
        # first test the end tag is there -- same as always:
        assert file_contents.startswith("<p")
    
        # but now the opening tag is far more complex
        # but it starts the same:
        assert file_contents.endswith("</p>")
    
        # order of the tags is not important in html, so we need to
        # make sure not to test for that
        # but each attribute should be there:
>       assert 'style="text-align: center"' in file_contents
E       assert 'style="text-align: center"' in '<p>\nA paragraph of text\n</p>'

test_html_render.py:256: AssertionError
-------------------------------------------------------------------- Captured stdout call --------------------------------------------------------------------
<p>
A paragraph of text
</p>
________________________________________________________________________ test_anchor1 ________________________________________________________________________

    def test_anchor1():
        """ Tests that you can add an anchor """
        a = A("http://google.com", "link to google")
    
        file_contents = render_result(a).strip()
        print(file_contents)
    
>       assert file_contents.startswith('<a ref="http://google.com">')
E       assert False
E        +  where False = <built-in method startswith of str object at 0x10400ced0>('<a ref="http://google.com">')
E        +    where <built-in method startswith of str object at 0x10400ced0> = '<a>link to google</a>'.startswith

test_html_render.py:394: AssertionError


students/BrianB/lesson04/mailroom2.py
```diff
@@ -5,23 +5,46 @@
 # Who: Brian Brumleve
```
nairraghav, line 5: It looks like you ran into a merge issue, instead of commenting out the changes that you don't want, you should just remove them from the file. Otherwise, this adds a bunch of extra commented out junk in your file which can affect readability and provides nothing

students/BrianB/lesson05/except_exercise.py
```diff
@@ -42,13 +50,24 @@
...
 langs = ['java', 'c', 'python']
 
+# <<<<<<< master
+try:
+    more_joke = more_fun(langs[0])
+except IndexError:
+    pass
+
+# =======
 more_joke = more_fun(langs[0])
```
nairraghav, line 60: This is already handled in your try/except above, I believe you meant to delete this?

students/BrianB/lesson05/except_exercise.py
```diff
@@ -42,13 +50,24 @@
...
+# >>>>>>> master
 
 try:
     not_more_joke = more_fun(langs[1])
 except SyntaxError:
     print('Run Away!')
 else:
+# <<<<<<< master
+# =======
     more_fun(langs[0])
```
nairraghav, line 70: Same thing here, I think you meant this previously but it's still here outside a try block. I think you meant to remove this as well.

students/BrianB/lesson06/mailroom4.py
```diff
@@ -0,0 +1,245 @@
...
+                       ))
+    print("You replied: ", answer)
+    answer_dict = {
+                   1: add_to_donations,
+                   2: print_donor_report,
+                   3: send_letter_to_all,
+                   4: quit
+                   }
+    ##########################################################
+    # QUESTION FOR RON: Why does parenthesis after .get work?
```
nairraghav, line 226: I believe we covered this during OH. Let me know if you still need more clarification on it!

students/BrianB/lesson06/mailroom4.py
```diff
@@ -0,0 +1,245 @@
...
+    answer_dict.get(answer, "Please input 1, 2, 3 or 4")()
+    return answer, answer_dict
+
+
+if __name__ == "__main__":
+    print("Welcome to the Mailroom!")
+
+    donor_data = donor_database
+
+    # main_menu()
```
nairraghav, line 236: Try to remove all commented out code in your PRs before submitting, it's a good habit to get into when writing code!

nairraghav:
It looks like you're having some merge issues here. Before actually deleting stuff away, I would advise that you review the code and make sure the files have the changes you want and then remove any uncommented code. Let me know if you want to spend some time walking through this together to figure out what the issues are.

students/BrianB/.gitignore
```diff
@@ -5,3 +5,14 @@ lesson04/Trigram text samples/
 lesson04/index_vs_slice.py
 lesson04/mutability.py
 lesson04/string_formatting_for_mailroom1.py
+lesson04/mailroom_for_lesson06.py
```
PythonCHB, line 8: you can put a dir in .gitignore, might be easier to put your notes/examples, etc all on one dir, and jsut .gitignore that.

but this is fine, too, if you don't mind maintaining it.

students/BrianB/lesson04/trigrams.py
```diff
@@ -1,4 +1,4 @@
-#! bin/usr/env python3
+#!/usr/bin/env python3
```
PythonCHB, line 1: better for print_trigram() to take the words as a parameter.

students/BrianB/lesson07/html_render.py
```diff
@@ -0,0 +1,225 @@
...
+            self.contents = []  # creates an empty list
+        else:
+            self.contents = [content]  # creates a list with contents
+            self.attribute = kwargs
+
+    def append(self, new_content):
+        self.contents.append(new_content)
+
+    def _open_tag(self):
+        tag = f"<{self.tag}>"
```
PythonCHB, line 33: the attributes need to be rendered here, in the open tag.

That's the cause of your test failures (only 2!)

students/BrianB/lesson07/html_render.py
```diff
@@ -0,0 +1,225 @@
...
+        return tag
+
+    def _close_tag(self):
```
PythonCHB, line 36: you aren't using this anywhere -- which is why you haven't noticed the missing /

students/BrianB/lesson07/html_render.py
```diff
@@ -0,0 +1,225 @@
...
+
+
+class Html(Element):
+    """
+    Framework for 'html' tag, inherits from Element.
+    Overrides render method and tag class attribute.
+    """
+    tag = 'html'
+
+    def render(self, out_file, indent="", ind_count=0):
```
PythonCHB, line 58: you want to either specify the indentation as a string: `indent=""` OR as a count: `ind_count=0`, but not both.

students/BrianB/lesson07/html_render.py
```diff
@@ -0,0 +1,225 @@
...
+        """
+        I was unable to follow the instructions beyond
+        mid-Step 4, this was borrowed from an alt source
```
PythonCHB, line 61: so this may be the source of your issue -- if you borrowed code that was using a different way to specify the indentation, you have a mish-mash

students/BrianB/lesson07/html_render.py
```diff
@@ -0,0 +1,225 @@
...
+        """
+        self.indent = indent
```
PythonCHB, line 63: Careful here -- the reason you are passing the indentation in to the render method is that it will be different depending on where the item is rendered -- it is needed within that function -- no need to store it in self.

students/BrianB/lesson07/html_render.py
```diff
@@ -0,0 +1,225 @@
...
+
+
+class Head(Element):
+    tag = "head"
+
+
+class Li(Element):
+    """
+    I was unable to follow the instructions beyond
+    mid-Step 4, this was borrowed from an alt source
```
PythonCHB, line 83: though this bit is pretty simple :-)

students/BrianB/lesson07/html_render.py
```diff
@@ -0,0 +1,225 @@
...
+    """
+    tag = 'ul'
+
+    def __init__(self, content=None, **kwargs):
+        """
+        I was unable to follow the instructions beyond
+        mid-Step 4, this was borrowed from an alt source
+        """
+        if content: raise TypeError
```
PythonCHB, line 100: when you raise an Exception like this, you want to give it a nice message:

raise TypeError("you can not add content to a UL element")

students/BrianB/lesson07/html_render.py
```diff
@@ -0,0 +1,225 @@
...
+
+        self.contents = []
```
PythonCHB, line 102: calling the super()'s __Init__ is a better idea here -- maybe it does a bit more? or will later.

students/BrianB/lesson07/html_render.py
```diff
@@ -0,0 +1,225 @@
...
+        self.attributes = kwargs
+
+
+class OneLineTag(Element):
+
+    def render(self, out_file):
+        # loop through the list of contents
```
PythonCHB, line 109: I don't see any looping here.

students/BrianB/lesson07/html_render.py
```diff
@@ -0,0 +1,225 @@
...
+        out_file.write("<{}>".format(self.tag))
+        out_file.write(self.contents[0])
+        out_file.write("</{}>\n".format(self.tag))
+
+    def append(self, content):
+        raise NotImplementedError
```
PythonCHB, line 115: This is not really the right Exception -- this is for when there is a feature that is not implimented, but maybe will be in the future.

TypeError is probably better here.

students/BrianB/lesson07/html_render.py
```diff
@@ -0,0 +1,225 @@
...
+
+    def __init__(self, level, content=None, **kwargs):
+        """
+        I was unable to follow the instructions beyond
+        mid-Step 4, this was borrowed from an alt source
+        """
+        try:
+            int(level)
+        except ValueError:
+            raise ValueError
```
PythonCHB, line 133: no need to catch an error simply to re-raise it.

though you might want to do that if you are going to add a message:

except ValueError:
            raise ValueError(the header level must be an integer between 1 and 6")

students/BrianB/lesson07/html_render.py
```diff
@@ -0,0 +1,225 @@
...
+    mid-Step 4, this was borrowed from an alt source
+    """
+    tag = 'a'
+
+    def __init__(self, link, content=None, **kwargs):
+        """
+        I was unable to follow the instructions beyond
+        mid-Step 4, this was borrowed from an alt source
+        """
+        if not (content and link): raise TypeError
```
PythonCHB, line 155: again, always add a message when you raise an Exception.

and do you need to have some content in an anchor tag? I'm not sure it it's required.

students/BrianB/lesson07/html_render.py
```diff
@@ -0,0 +1,225 @@
...
+    """
+    I was unable to follow the instructions beyond
+    mid-Step 4, this was borrowed from an alt source
+    """
+    def __init__(self, content=None, **kwargs):
+        """
+        I was unable to follow the instructions beyond
+        mid-Step 4, this was borrowed from an alt source
+        """
+        if content: raise TypeError
```
PythonCHB, line 171: if it can't take any content, you can not have it in the parameters at all:

`def __init__(self, **kwargs):`

students/BrianB/lesson07/html_render.py
```diff
@@ -0,0 +1,225 @@
...
+        mid-Step 4, this was borrowed from an alt source
+        """
+        raise NotImplementedError
+
+    def render(self, out_file, indent="", ind_count=0):
+        """
+        I was unable to follow the instructions beyond
+        mid-Step 4, this was borrowed from an alt source
+        """
+        self.indent = indent * ind_count
```
PythonCHB, line 186: you don't need to set it in self.

and in this context, you don't want the ind_count multiplied by the indent passed in. If you are using a count, then it would be:

cur_indent = self.indent * ind_count

where self.indent is a class attribute of Element -- and does not change.

students/BrianB/lesson07/html_render.py
```diff
@@ -0,0 +1,225 @@
...
+    mid-Step 4, this was borrowed from an alt source
+    """
+    tag = 'br'
+
+    def __init__(self, content=None, **kwargs):
+        """
+        I was unable to follow the instructions beyond
+        mid-Step 4, this was borrowed from an alt source
+        """
+        if kwargs: raise TypeError
```
PythonCHB, line 215: raise an Exception if there are kwargs, then pass them on? no need -- if they are sure to be empty then you don't need to pass them along.

PythonCHB (changes requested):
you got a good start here.

but it needs some clean up and adding rendering the attributes.

I hope class this week clarified some of this for you. IF not -- look at my solutions, and come to office hours if you can.
I'll give partial credit

PythonCHB:
The merge cruft is mostly in older stuff. Still good to clean it up, but I'll merge for now, so it will be clear what's new the next time.

PR #148, opened 2019-05-25
Adding HTML render

ervbrian:
Test results:

```
This only tests that it can be initialized with and without ... ok
This tests that you can append text ... ok
Tests whether the Element can render two pieces of text ... ok
Tests whether the Element can render two pieces of text ... ok
test_html_render.test_html ... ok
test_html_render.test_body ... ok
test_html_render.test_p ... ok
tests that you can add another element and still render properly ... ok
test_html_render.test_head ... ok
test_html_render.test_title ... ok
test_html_render.test_title_append ... ok
test_html_render.test_attributes ... ok
a simple horizontal rule with no attributes ... ok
a horizontal rule with an attribute ... ok
a href test ... ok
a href test ... ok
Unordered list test ... ok
Tests that the indentation gets passed through to the renderer ... ok
The contents in a element should be indented more than the tag ... ok
make sure multiple levels get indented fully ... ok
ok

----------------------------------------------------------------------
Ran 21 tests in 0.026s

OK
```

### DIFF
--- a/students/BrianE/session07/html_render.py
+++ b/students/BrianE/session07/html_render.py
@@ -1,0 +1,146 @@
+#!/usr/bin/env python3
+
+"""
+A class-based system for rendering html.
+"""
+
+class TextWrapper:
+    """
+    A simple wrapper that creates a class with a render method
+    for simple text
+    """
+    def __init__(self, text):
+        self.text = text
+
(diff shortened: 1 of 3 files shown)


nairraghav:
Excellent job Brian! Everything looks good 🎉

PR #149, opened 2019-05-27
session07 : HTML Rendering project and other excercises

JoeNunnelley:


students/JoeNunnelley/session07/dog.py
```diff
@@ -0,0 +1,50 @@
+#! /usr/bin/env python3
```
nairraghav, line 1: Great example, nice job covering it in class!

students/JoeNunnelley/session07/html_render.py
```diff
@@ -0,0 +1,242 @@
...
+                else:
+                    new_content.depth = self.depth + 1
+                    self.children.append(new_content)
+            elif isinstance(new_content, dict):
+                self.attributes.append(new_content['attributes'])
+                self.text.append(new_content['text'])
+                self.children = new_content['children']
+            elif isinstance(new_content, str):
+                self.text.append(new_content)
+            elif new_content is None:
```
nairraghav, line 59: I don't believe you need this since it will never enter the if statement if new_content is None

students/JoeNunnelley/session07/html_render.py
```diff
@@ -0,0 +1,242 @@
...
+        """
+        This function renders children that are not at the root of the document.
+        """
+        children_text = ''
+        attribs = ''
+        for child in children:
+            if child.attributes and child.attributes[0]:
+                attribs = ' ' + ''.join(child.attributes)
+
+            start_tag = "\n{}<{}{}>".format((child.depth * child.indent), child.tag, attribs)
```
nairraghav, line 73: Completely works but consider using f strings for more readability!

nairraghav:
Great work Joe!

PR #151, opened 2019-05-28
Finished Lesson 7

Charlie-England:
I've completed as best I could.
I had a hard time on step 5 and the self._open_tag() methods so going over this in class might be helpful. It may be something simple but I had a hard time understanding what was being asked.
I dont know where self._open_tag() is supposed to go.
I also had trouble with the A Class (href), it's not displaying correctly and I had to solve it in a way that I'm not fond of. I'll try to work on that later but as of right now it is not displaying correctly.

### DIFF
--- a/students/charlie_england/session07/html_render.py
+++ b/students/charlie_england/session07/html_render.py
@@ -1,0 +1,51 @@
+#!/usr/bin/env python3
+
+"""
+A class-based system for rendering html.
+"""
+
+# This is the framework for the base class
+class Element(object):
+    tag = "html"
+
+    def __init__(self, content=None):
+        self.content_lst = [f"<{self.tag}>",]
+        if content:
+            self.content_lst.append(content)
+
+    def append(self, *new_content):
+        for item in new_content:
+            self.content_lst.append(item)
+        
+
+    def render(self, out_file):
+        for line in self.content_lst:
+            try:
+                line.render(out_file)
+            except AttributeError:
+                out_file.write(f"{line}\n")
+        out_file.write(f"</{self.tag}>\n")
+            
+class Body(Element):
+    tag = "body"
+
+class Html(Element):
+    tag = "html"
+
+class P(Element):
+    tag = "p"
+
+class Head(Element):
+    tag = "head"
+
+class OneLineTag(Element):
+    def render(self, out_file):
+        for line in self.content_lst:
+            try:
+                line.render(out_file)
+            except AttributeError:
+                out_file.write(f"{line}")
+        out_file.write(f"</{self.tag}>")
+
+class Title(OneLineTag):
+    tag = "title"
--- a/students/charlie_england/session07/html_render.py
+++ b/students/charlie_england/session07/html_render.py
@@ -8,10 +8,14 @@ A class-based system for rendering html.
 class Element(object):
     tag = "html"
 
-    def __init__(self, content=None):
+    def __init__(self, content=None,**attributes):
         self.content_lst = [f"<{self.tag}>",]
+        self.__dict__.update(attributes)
         if content:
             self.content_lst.append(content)
+        if attributes:
+            self.content_lst.pop(0)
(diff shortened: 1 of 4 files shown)


students/charlie_england/session07/html_render.py
```diff
@@ -0,0 +1,109 @@
...
+"""
+A class-based system for rendering html.
+"""
+
+# This is the framework for the base class
+class Element(object):
+    tag = "html"
+
+    def __init__(self, content=None,**attributes):
+        self.content_lst = [f"<{self.tag}>",]
```
PythonCHB, line 12: hmm, I haven't seen this approach before.

But I think it's better to wait and add the tag in the render method -- this way, it is very fagile if the content os manipulated later on: the element is created without content, and content added later, or content removed (not part of the spec, but still...), etc.

students/charlie_england/session07/html_render.py
```diff
@@ -0,0 +1,109 @@
...
+        self.__dict__.update(attributes)
```
PythonCHB, line 13: careful here! this is kinda black magic! Yes, it can work, but you have updated the namespace of the entire instance -- now your attributes are tangled up with the regular instance attributes: `self.content_lst` for example. And there may be more in the future.

Better to keep the tag attributes in their own separate dict.

students/charlie_england/session07/html_render.py
```diff
@@ -0,0 +1,109 @@
...
+        Element.render(self,out_file)
+
+class P(Element):
+    tag = "p"
+
+class Head(Element):
+    tag = "head"
+
+class OneLineTag(Element):
+    def render(self, out_file):
```
PythonCHB, line 50: I'm seeing a lot of repeated code here: DRY!

students/charlie_england/session07/html_render.py
```diff
@@ -0,0 +1,109 @@
...
+                line.render(out_file)
+            except AttributeError:
+                out_file.write(f"{line}")
+        out_file.write(f"</{self.tag}>\n")
+
+class Title(OneLineTag):
+    tag = "title"
+
+class SelfClosingTag(Element):
+    def __init__(self, content=None,**kwargs):
```
PythonCHB, line 62: why have a content parameter at all?

students/charlie_england/session07/html_render.py
```diff
@@ -0,0 +1,109 @@
...
+        if content is not None:
+            raise TypeError("SelfClosingTag can not contain any content")
+        super().__init__(content=content,**kwargs)
+
+    def render(self, out_file):
+        if len(self.content_lst[0]) > 4:
```
PythonCHB, line 68: huh? what are you looking for here? seems very fragile!

students/charlie_england/session07/html_render.py
```diff
@@ -0,0 +1,109 @@
...
+class A(OneLineTag):
+    tag = "a"
+
+    def __init__(self,link=None,content=None):
+        self.content = content
+        self.link = link
+        super().__init__(content=content)
+
+    def render(self, out_file):
+        out_file.write(f'<{self.tag} href="{self.link}">{self.content}</{self.tag}>')
```
PythonCHB, line 92: a link is really like any other attribute -- if you add it to the attribute collection, then the same code can be used to render the link as the other attributes.

students/charlie_england/session07/test_html_render.py
```diff
@@ -0,0 +1,349 @@
...
+def test_ele_attributes():
+    attributes = {"id":"TheList","style":"line-height:200%"}
+    e = Body("here is some body text",**attributes)
+
+    file_contents = render_result(e)
+    assert "here is some body text" in file_contents
+    assert 'id="TheList"' in file_contents
+
+#I had a lot of trouble with this part and did not understand the tutorial:
+#If you put the code to render the opening (and closing) tags in it’s own method, then we can call that method from multiple render methods, something like
```
PythonCHB, line 231: in your case, you rendered the opening tag right away when the element was created -- so yes, this doesn't make sense in that context.

Better to render to create the tag at render time -- then you can have a method that just does that. And override it when you want to do it differently.

PythonCHB (changes requested):
nice work! you got most of it.

Some refactoring is in order, and you still need to add indentation. So I'm going to hold of on giving credit 'till you get to that.

PR #154, opened 2019-05-29
assignments from 3, 5, 6

trayffs:
From lesson3 I've included slicing lab, and string formatting lab.
From lesson5 I've included exercise exceptions (I've skipped mailroom part 3 in favor of working on mailroom part 4)
from lesson6 I've included mailroom part 4 and mailroom test.


students/TylerRay/session05/except_exercise.py
```diff
@@ -0,0 +1,45 @@
...
+# If there are no exceptions, call the more_fun function with the last
+# language in the list
+
+# Finally, while still in the try/except block and regardless of whether
+# there were any exceptions, call the function last_fun with no
+# parameters. (pun intended)
+
+langs = ['java', 'c', 'python']
+
+more_joke = more_fun(langs[1])
```
PythonCHB, line 45: You still need this last bit ....

one more way to handle exceptions.

students/TylerRay/session3/slicing_lab.py
```diff
@@ -0,0 +1,45 @@
+
+test_string = "this is my chosen test string."
+test_tuple = ("thing 1", "thing 2", 54, 40.1, "last thing")
```
PythonCHB, line 3: nice use of multiple types in a tuple -- easy to see the results.

students/TylerRay/session3/slicing_lab.py
```diff
@@ -0,0 +1,45 @@
...
+test_list = [12, 45, 31, 12, 56, 88, 97, 4, 23, 63, 34, 90]
+
+
+def string_repeat(str):
+    print("test 1: ",str)
+
+
+def first_and_last_swap(str):
+    print("test 2: ",str[-1:], str[1:-1], str[0:1])
```
PythonCHB, line 12: you've got the slicing right, but this is printing the three parts separately, rather than putting them together.

And you should return the result, rather than just printing it.

Since you are doing this now -- you could even test all these...

students/TylerRay/session3/strformat_lab.py
```diff
@@ -1,17 +1,51 @@
-tpl1 = ( 2, 123.4567, 10000, 12345.67)
+tpl1 = (2, 123.4567, 10000, 12345.67)
 
+# Task 1
 def printTuple(tpl1):
     v1 = "file_{0:03}: ".format(tpl1[0])
     v2 = "{0:.2f}".format(tpl1[1])
     v3 = "{:.2e}".format(tpl1[2])
     v4 = "{0:.2e}".format(tpl1[3])
     return f"{v1} {v2} {v3} {v4}"
```
PythonCHB, line 9: You can do all this in one format string:

"{0:.2f} {1:.2e} {1:.2e} {3:.2e} ".format(tp[0], tp[1], tp[2], tp[3])

or

"{0:.2f} {1:.2e} {1:.2e} {3:.2e} ".format(*tp)

students/TylerRay/session3/strformat_lab.py
```diff
@@ -1,17 +1,51 @@
...
+# Task 2
+print("Task 2")
+print("file_{:03}:  {:.2f} {:.2e} {:.2e}".format(*tpl1))
+
+
+# Task 3
+tpl2 = 12, 45, 22, 587, 97, 43, 32,
+
+def formatter(tpl2):
+    return (f"the {len(tpl2)} numbers are:" + " {:d},"*(len(tpl2)-1)) + " {:d}"
```
PythonCHB, line 24: nice!

students/TylerRay/session6/mailroom_4.py
```diff
@@ -0,0 +1,225 @@
...
+
+
+def find_donor(name):
+    '''
+    Find a donor in donor_db
+    :param: the name of the donor
+    :returns: the donation (value) data, None if not in donor_db
+    '''
+
+    key = name.strip()
```
PythonCHB, line 42: it looks like you have lower-cased the name to use as a key -- so you should probably call lower() here too.

students/TylerRay/session6/mailroom_test.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def test_report():
+    '''
+    Checks to see if the gen_donor_report() outputs as expected
+    :return:
+    '''
+    result = mr.gen_donor_report()
+
+    assert "Vladmir Putin               $1500000.00           3   $  500000.00" in result
```
PythonCHB, line 31: it would be good to add a test that it is sorted correctly, too.

PythonCHB (approved):
nice work !

I"m holding off credit for the exception exercise, it's not quite done.

But credit for everything else.

NOTE: that seems like not very many tests for mailroom -- if you have no other functions that don't have input() in them, maybe more refactoring to get as much logic as possible away from the code that is asking for user input.

You'll have one more chance with an OO mailroom -- so see what you can do then

PR #156, opened 2019-05-31
Adds slicing lab

jamessward:
feedback always appreciated!


PythonCHB:
very nice!

We're getting to the end of the class -- you've got a fair bit to catch up on.  

I suggest a full-on tested mailroom with dicts and exeptions -- we can give you credit for the first 4 versions all at once.

PR #158, opened 2019-06-01
here is the html render

gjkim44:
the code orginally was written half way to step 5.  15 hours later.  Very hard assignment for a class that was not taught by you Chris.  Brian did his best, but I honestly was lost

### DIFF
--- a/students/GKim/lesson07/html_render.py
+++ b/students/GKim/lesson07/html_render.py
@@ -1,0 +1,26 @@
+#!/usr/bin/env python3
+
+"""
+A class-based system for rendering html.
+"""
+
+
+# This is the framework for the base class
+class Element:
+
+    tag = "html"
+
+    def __init__(self, content=None):
+       self.contents = [content]
+
+    def append(self, new_content):
+        self.contents.append(new_content)
+        
+
+    def render(self, out_file):
+        out_file.write("<{}>\n".format(self.tag))
+        for content in self.contents:
+            out_file.write(content)
+            out_file.write("\n")
+        out_file.write("</{}>\n".format(self.tag))
+
--- a/students/GKim/lesson07/html_render.py
+++ b/students/GKim/lesson07/html_render.py
@@ -2,25 +2,301 @@
 
 """
 A class-based system for rendering html.
+This is someone elses code...very difficult assignment and I did
+I did not learn from it
 """
 
 
 # This is the framework for the base class
-class Element:
+class Element():
+    """
+    Framework for a basic element in HTML code.
+    Class attributes:
+    tag:     Tag used for rendering html contents.  Element class is set to html.
+    indent:  The level of indentation for the current element.
+    """
+    tag = 'html'
+    indent = ''
 
-    tag = "html"
+    def __init__(self, content=None, **kwargs):
+        """
+        __init__ method for Element class.
+        :param content: Content that is to be added to the instance content and later rendered.
+        :param kwargs:  Used for passing in tag attributes.
+        Attributes:
+        contents    The content that is to be rendered into HTML format.
+        """
+        self.attributes = kwargs
+        self.contents = [content] if content else []
 
-    def __init__(self, content=None):
-       self.contents = [content]
+    def _open_tag(self):
+        """ Returns the opening tag for the current element """
+        return f'<{self.tag}>'
+
+    def _close_tag(self):
+        """ Returns the closing tag for the current element """
+        return f'</{self.tag}>'
 
     def append(self, new_content):
+        """
+        Appends new_content to the instance attribute content
+        :param new_content: String that is to be appended to the instance content.
+        """
         self.contents.append(new_content)
-        
 
-    def render(self, out_file):
-        out_file.write("<{}>\n".format(self.tag))
+    def render(self, out_file, indent="", ind_count=0):
+        """
+        Recursively renders the instance attribute content into HTML format code.
+        :param out_file: Destination for rendered text.
+        :param indent: The specified indentation level for elements.
+        :param ind_count: How many times indent should be applied to the element.
+        """
+        self.indent = indent * ind_count
+
+        out_file.write(f'{self.indent}{self._open_tag()[:-1]}')
+        for key, value in self.attributes.items():
+            out_file.write(f' {key}="{value}"')
+        out_file.write(f'{self._open_tag()[-1:]}\n')
+
         for content in self.contents:
-            out_file.write(content)
-            out_file.write("\n")
-        out_file.write("</{}>\n".format(self.tag))
+            if hasattr(content, 'render'):
+                content.render(out_file, indent, ind_count + 1)
+            else:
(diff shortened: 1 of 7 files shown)


PythonCHB:
Got it George, we'll take a look soon. And hopefully Office hours this weekend can help.

Sorry about the timing -- I know that this is a tough one, so try to really get it going in class before hand, but I didn't think to suggest that to Bryan.

Also -- if the tutorial is unclear, please do let us know where it needs more exposition, and if possible, how to clarify it.

students/GKim/lesson07/html_render.py
```diff
@@ -0,0 +1,302 @@
...
+        self.contents.append(new_content)
+
+    def render(self, out_file, indent="", ind_count=0):
+        """
+        Recursively renders the instance attribute content into HTML format code.
+        :param out_file: Destination for rendered text.
+        :param indent: The specified indentation level for elements.
+        :param ind_count: How many times indent should be applied to the element.
+        """
+        self.indent = indent * ind_count
```
PythonCHB, line 54: if you are using this approach -- passing in the number of indentation levels, then yhou dont want to also pass in the indent -- you should use self.indent for that.

PythonCHB, line 54: you do not want to be re-setting self.indent here --that is a class attribute that is suposed to hold a single indentation level.

so this should be:

cur_indent = self.indent * ind_count

students/GKim/lesson07/html_render.py
```diff
@@ -0,0 +1,302 @@
...
+
+        out_file.write(f'{self.indent}{self._open_tag()[:-1]}')
```
PythonCHB, line 56: no need for formatting when it's jsut two things:

out_file.write(cur_indent + self._open_tag()[:-1])

but why are you slicing off the end of the open tag???

students/GKim/lesson07/html_render.py
```diff
@@ -0,0 +1,302 @@
...
+        for key, value in self.attributes.items():
+            out_file.write(f' {key}="{value}"')
```
PythonCHB, line 58: this code should be in the _open_tag method -- that's kind of the point of having it -- so it can build the whole open tag for you.

students/GKim/lesson07/html_render.py
```diff
@@ -0,0 +1,302 @@
...
+        out_file.write(f'{self._open_tag()[-1:]}\n')
+
+        for content in self.contents:
+            if hasattr(content, 'render'):
+                content.render(out_file, indent, ind_count + 1)
```
PythonCHB, line 63: now you are passing managing teh level of indention in two ways -- could be prone to error:

content.render(out_file, ind_count + 1)

students/GKim/lesson07/html_render.py
```diff
@@ -0,0 +1,302 @@
...
+            else:
+                out_file.write(f'{self.indent}{indent}{content}\n')
```
PythonCHB, line 65: and this can be:

out_file.write(cur_indent + content + "\n")

students/GKim/lesson07/html_render.py
```diff
@@ -0,0 +1,302 @@
...
+
+        out_file.write(f'{self.indent}{self._close_tag()}\n')
```
PythonCHB, line 67: and then:

out_file.write(f'{cur_indent}{self._close_tag()}\n')

Key here is to think about what each of these names represetn, so you can know where to keep them:

"self" is for keeping teh current state of object, so that multiple methods all have access. So ony store things there that will be needed by other methods.

in this case:

self.indent is a class attribute, that represents a single level of indentation for ALL Elements -- youdo'nt want to change that.

ind_count is a function local variable that indicates how many levels of indentation this particular element needs to be indented while it is being rendered.

cur_indent (= self.indent * ind_count)is the current level of indentation for this element, while it is being rndered -- so it is only needed inside the render function, so no need to assign to self.

-CHB

students/GKim/lesson07/html_render.py
```diff
@@ -0,0 +1,302 @@
...
+    tag = 'html'
+
+    def render(self, out_file, indent="", ind_count=0):
+        """
+        Adds <!DOCTYPE html> tag to output then calls super().render after running
+        :param out_file: Destination for rendered text.
+        :param indent: The specified indentation level for elements.
+        :param ind_count: How many times indent should be applied to the element.
+        """
+        self.indent = indent
```
PythonCHB, line 84: again, just pass ind_count through -- no need to mess with self.indent

students/GKim/lesson07/html_render.py
```diff
@@ -0,0 +1,302 @@
...
+        Renders tag and attributes on a single line
+        Overrides Element.render
+        :param out_file: Destination for rendered text.
+        :param indent: The specified indentation level for elements.
+        :param ind_count: How many times indent should be applied for the element.
+        """
+        self.indent = indent * ind_count
+
+        out_file.write(f'{self.indent}{self._open_tag()[:-1]}')
+        for key, value in self.attributes.items():
```
PythonCHB, line 165: DRY! This is why you want this code int he _open_tag method :-)

students/GKim/lesson07/html_render.py
```diff
@@ -0,0 +1,302 @@
...
+        :param kwargs:  Used for passing in tag attributes.
+        """
+
+        # Verifies function called with interger value for header level as first parameter.
+        # If it was forgotten, then contents will likely be the first passed value.
+        # This 'should' then fail if casted to an int, and raises a ValueError
+        try:
+            int(level)
+        except ValueError:
+            raise ValueError
```
PythonCHB, line 193: ass a nice message if you re-raise -- other wise, no need to catch and re-raise the same Exception.

PythonCHB (commented):
this code is looking familiar -- did you work with Miguel on this? if so, take alook at the comments on his code, too :-)

students/GKim/lesson08/test_circle.py
```diff
@@ -0,0 +1,146 @@
...
+    c1 = Circle(1)
+    c2 = Circle(2)
+
+    assert c1 < c2
+
+def test_sort_circle():
+
+    c_list = [Circle(2), Circle(7), Circle(3)]
+
+    c_list.sort(key=Circle.sort_key)
```
nairraghav, line 101: Just by implementing the __lt__, you should be able to sort without giving a key

students/GKim/lesson08/test_circle.py
```diff
@@ -0,0 +1,146 @@
...
+    assert s1.area ==  201.06192982974676
+
+
+def test_sphere():
+    s = Sphere(5)
+
+    assert s.radius == 5
+    assert s.diameter == 10
+
+# def test_sphere_area():
```
nairraghav, line 119: Since you've implemented this, you can have a test validate the area

nairraghav:
Nice job with the Circle assignment George! I'll merge this in but take a look at what Chris had mentioned for the HTML Render assignment

PR #159, opened 2019-06-01
session08 homework

JoeNunnelley:
- sparse_array work
- circle/sphere work using magic methods

### DIFF
--- a/students/JoeNunnelley/session08/circle.py
+++ b/students/JoeNunnelley/session08/circle.py
@@ -116,6 +116,15 @@ class Circle:
         (new_self, new_other) = self.harmonize_types(other)
         return Circle(new_self ** new_other)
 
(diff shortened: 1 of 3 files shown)


students/JoeNunnelley/session08/circle.py
```diff
@@ -0,0 +1,220 @@
...
+        return math.pi * (self.radius * self.radius)
+
+
+    @classmethod
+    def from_diameter(cls, diameter):
+        """initialize a circle using diameter"""
+        return cls(diameter / 2)
+
+
+    def harmonize_types(self, other):
```
nairraghav, line 66: Nice use here to avoid repeating yourself

nairraghav:
Nice job Joe!

PR #162, opened 2019-06-02
PR For Last Two Weeks

miotke:
Includes week 7 and week 8


students/AndrewMiotke/session07/HTML_Renderer/html_render.py
```diff
@@ -0,0 +1,129 @@
+#!/usr/bin/env python3
+
+"""
+A class-based system for rendering html.
+"""
+
+# This is the framework for the base class
+class Element: # Did we need the () in Python3+ ?
```
nairraghav, line 8: Nope! I believe you'll only have to specify the parameters if you use them (inheritance)

students/AndrewMiotke/session07/HTML_Renderer/html_render.py
```diff
@@ -0,0 +1,129 @@
...
+        else:
+            self.contents = []
+
+
+    def append(self, new_content):
+        self.contents.append(new_content)
+
+
+    def make_tags(self):
+        attributes = " ".join([f'{key}="{val}' for key, val in self.attributes.items()])
```
nairraghav, line 27: This is throwing me an error when I try to run this, are you able to run this properly?

students/AndrewMiotke/session07/HTML_Renderer/html_render.py
miotke: 👍 cool, that's what I thought but figured I should ask.

students/AndrewMiotke/session07/HTML_Renderer/html_render.py
miotke: What error are you getting? I'm not receiving an error but I'm also not getting any output 🤔

miotke: Found the error, 
```
Traceback (most recent call last):
  File "run_html_render.py", line 46, in <module>
    render_page(page, "test_html_output1.html")
  File "run_html_render.py", line 27, in render_page
    page.render(f)
  File "/Users/andrewmiotke/GitHub/SP_2019_210A_classroom/students/AndrewMiotke/session07/HTML_Renderer/html_render.py", line 44, in render
    stuff.render(out_file, current_indent + self.indent)
AttributeError: 'str' object has no attribute 'render'
```

I can't seem to find the reason why. My best guess is that string doesn't know of render but I'm not sure why.

nairraghav: It looks like the issue here is that you're running the render on your self.contents. The render is a function of the WebElement not a string

nairraghav: It looks like the very first item in your "stuff" is a String. This needs to be an element.

nairraghav:
I'm going to go ahead and merge this in to get your Circle changes. Please take a look at that error and try to get past that String render issue in a future PR! 

I believe once you have that resolved, the rest should be working.

miotke:
Thanks Ron, working on those suggestions now and will update this code and add in session 10 in another PR

PR #163, opened 2019-06-02
Circle Class and Tests for Circle Class

roviram:
Here is my circle class and tests for my circle class assignment for lesson08! Thank you for reviewing!

### DIFF
--- a/students/roviram/lesson08/test_circle_class.py
+++ b/students/roviram/lesson08/test_circle_class.py
@@ -1,0 +1,155 @@
+# Dev: Miguel Rovira-Gonzalez
+# Lesson08: Circle Assignment
+
+import pytest
+import math
+from circle_class import Circle, Sphere
+
+
+def test_init():
+    c = Circle(10)
+    assert c.radius == 10
+
+
+def test_no_radius():
+    with pytest.raises(TypeError):
+        c = Circle()
+
+
+def test_diameter():
+    c = Circle(10)
+    assert c.diameter == 20
+
+
+def test_diameter_change_radius():
+    c = Circle(10)
+    assert c.diameter == 20
+    c.radius = 4
+    assert c.radius == 4
+    assert c.diameter == 8
+
+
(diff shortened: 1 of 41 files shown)


students/roviram/Lesson04/trigrams.py
```diff
@@ -0,0 +1,103 @@
...
+
+def read_file(filename):
+    """
+    Read the file and turn it into a list of words
+    :param filename:
+    :return: list of words in the file
+    """
+    words = []
+    # Code wasn't working unless I put the full file path in
+    with open(r"C:\Users\roviram\Python_210\SP_2019_210A_classroom\students\roviram\Lesson04\sherlock.txt") as text:
```
nairraghav, line 37: If you run this from your lesson04 directory, you shouldn't need to specify this entire path

students/roviram/lesson08/test_circle_class.py
```diff
@@ -0,0 +1,155 @@
...
+    assert str(c) == "Circle with radius: 20"
+
+
+def test_rep():
+    c = Circle(10)
+    assert repr(c) == "Circle10"
+
+
+def test_set_area():
+    c = Circle(5)
```
nairraghav, line 58: You won't need this error here since you will never throw an AttributeError for Circle. The area method is implemented here.

students/roviram/lesson08/test_circle_class.py
```diff
@@ -0,0 +1,155 @@
...
+
+def test_instance_equality():
+    c1 = Circle(10)
+    c2 = c1
+    assert c1 == c2
+
+
+def test_sort():
+    circle_list = [Circle(20), Circle(10), Circle(15), Circle(5)]
+    circle_list.sort(key=Circle.sort_key)
```
nairraghav, line 106: Since you specified the __lt__ method, you don't need to specify a sort key

nairraghav:
Nice job Miguel!

PR #165, opened 2019-06-04
completed circle.py and test_circle.py

Charlie-England:
should all be set, let me know if I missed something


nairraghav:
Nice job with the Circle assignment!

PR #169, opened 2019-06-05
Adds circle exercise

jamessward:


students/jsward/session08/circle.py
```diff
@@ -0,0 +1,71 @@
+import math
+
+
+class Circle:
+    def __init__(self, radius):
+        self.radius = radius
+
+    def __add__(self, other):
```
nairraghav, line 8: Just a note, it may be a good idea to validate this other item is of the same class, otherwise you could run into errors :)

nairraghav:
Nice job James!

PR #171, opened 2019-06-07
lesson 7 files

trayffs:
The HTML render doesn't really work. I tried to go through the tutorial but kept getting lost around step 3. I looked at other peoples solutions which was insightful. I also tried to straight up copy your work to try to get it working but never got it to function as expected. I'm turning in what I have because I need to move on to other assignments!


students/TylerRay/session7/run_html_render.py
```diff
@@ -0,0 +1,231 @@
...
+page = hr.Element()
+
+page.append("Here is a paragraph of text -- there could be more of them, "
+            "but this is enough  to show that we can do some text")
+
+page.append("And here is another piece of text -- you should be able to add any number")
+
+render_page(page, "test_html_output1.html")
+
+The rest of the steps have been commented out.
```
nairraghav, line 48: These should be commented out or removed

students/TylerRay/session7/html_render.py
```diff
@@ -0,0 +1,183 @@
+#!/usr/bin/env python3
```
nairraghav, line 1: There are a couple of Syntax errors here which I was able to quickly resolve and run. I will give partial credit for what you've completed thus far, the render isn't fully working for me.

nairraghav:
Hi Tyler, I will give partial credit for the HTML render assignment thus far. I have given credit for the Circle assignment.

PR #173, opened 2019-06-08
Ready to be reviewed

FredBallyns:
Had a bit of trouble when tried to remake original selection table as a switch dictionary. It kept auto running all choices when dictionary was defined. Am I suppose to drop the "()" from the end of the function and somehow add them back?

### DIFF
--- a/students/FredB/session06/mailroom6.py
+++ b/students/FredB/session06/mailroom6.py
@@ -1,0 +1,137 @@
+#Mailroom Part2
+#FredBallyns
+#Session06
+
+import sys, os
+
+donors = {"Fred Flintstone": [100, 200, 50],
+        "James Bond": [0,0,7],
+        "Lex Luthor": [5000,],
+        "Harambe": [3, 2, 1],
+        "Herman Cain": [9, 9, 9]
+		}
+
+def clear_screen():
+    #Windows 'nt'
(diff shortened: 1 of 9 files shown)


nairraghav:
> Had a bit of trouble when tried to remake original selection table as a switch dictionary. It kept auto running all choices when dictionary was defined. Am I suppose to drop the "()" from the end of the function and somehow add them back?

Yep! In the dict, just have the name with out the parenthesis, because that will call the function and map the result to the key.

When the user picks the index in the dict, you can call it like this: 
dict_name\[choice\]()

students/FredB/session06/mailroom6.py
```diff
@@ -0,0 +1,143 @@
...
+    elif not repeat_donor(donor):
+        new_donor = input("Donor not found. Is this a new donor? (y/n): ")
+        if new_donor[0].lower() == "y":
+            donors[donor]=[]
+        else:
+            return
+    else:
+        print("Donation again :)")
+    donation = -1
+    while valid_donation(donation)==False:
```
nairraghav, line 61: you can write this as:
while not valid_donation(donation):

nairraghav:
Excellent job Fred, I've given you credit for all the assignments here.

PR #174, opened 2019-06-09
Adds Mailroom OO

jamessward:
Certainly interested in feedback on the final.  Let me know if anything is missing or you find bugs.

### DIFF
--- a/students/jsward/session09/mailroom_oo/cli_main.py
+++ b/students/jsward/session09/mailroom_oo/cli_main.py
@@ -1,0 +1,185 @@
+#!/usr/bin/env python
+
+from donor_models import Donor, DonorCollection
+from operator import itemgetter
+
+############################
+# DONOR MANAGEMENT FUNCTIONS
+############################
+
+
+def add_donation():  # Tested
+    """
+    Adds a donation to the specified donor.  Prints a thank you letter to stdout.
+    :return: None
+    """
+    donor = find_donor()
+    while True:
+        donation_input = input("Input {}'s donation amount in dollars\n".format(donor.name))
+        try:
+            donor.add_donation(donation_input)
+            print(donor.thank_you_letter)
+            return
+        except ValueError:
+            print("{} is not a valid donation amount.".format(donation_input))
+            continue
+
+
+def add_donor():  # Tested
+    """
+    Adds a donor to the donor collection
+    :return: None
+    """
+    donor_valid = False
+    while not donor_valid:
+        try:
+            user_input = str(input("Provide the name to add to the donor list\n"))
+        except TypeError:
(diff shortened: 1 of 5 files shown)


students/jsward/session09/mailroom_oo/cli_main.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+def find_donor():  # Tested
+    """
+    Prompts user for a donor name.  Locates the corresponding Donor object in the DonorCollection
+    :return: Donor object
+    """
+    while True:
+        user_input = str(input("Provide the name of the donor.\n"))
+        for donor in donor_collection.donors:
```
nairraghav, line 97: It would have been good to implement a function in your donor_collection object to do this search for you, rather than have the cli_main handle this logic

students/jsward/session09/mailroom_oo/donor_models.py
```diff
@@ -0,0 +1,130 @@
...
+        if len(self.donations) == 0:
+            letter_contents = "Dear {}, we look forward to your first donation!".format(self.name)
+        else:
+            letter_contents = "Dear {}, thank you for the very generous donation of ${:.2f}.  " \
+                              "Rest assured it will be put to good use helping those in need*\n" \
+                              "    *After deducting 98% for management overhead, of course ;).".format(self.name,
+                                                                                                   self.donations[-1])
+        return letter_contents
+
+    def add_donation(self, amount):
```
nairraghav, line 60: Nice job putting this logic in here!

students/jsward/session09/mailroom_oo/donor_models.py
```diff
@@ -0,0 +1,130 @@
...
+            raise ValueError
+        self.donations.append(amount)
+
+
+class DonorCollection:
+    def __init__(self):
+        """
+        Initializes an instance of the DonorCollection class
+        """
+        self.donors = []
```
nairraghav, line 80: A list is a great start but a dict would have been a good option to use here. If you have a lot of donors in this list, you will have to iterate through the list every single time you want to find a donor. If you use a dict, that becomes much much faster.

nairraghav: In class, we went over creating a normalized name. In practice, that would be the key used for the donor object in this dict.

nairraghav:
Nice job James, I think you did a good job with the OOP Mailroom. When coding with OOP goals, try to think about where your logic would go. For example: any logic involving the donor object would go in the donor object while anything involving the collection itself would go in the collection object. I think using the list is an acceptable way to create donorcollection but in the future, consider the other data structures we covered! 

Nice work :)

PR #177, opened 2019-06-11
session09 oo mailroom

JoeNunnelley:
the oo mailroom implementation with donations, donor and donorcollection classes along with associated tests.


students/JoeNunnelley/session09/cli_main.py
```diff
@@ -0,0 +1,300 @@
...
+The Mailroom Program : Part 4 : Object Oriented
+Author : Joe Nunnelley
+Description:
+    A small command line based mailroom application designed to thank donors.
+"""
+import datetime
+import os
+import sys
+import tempfile
+import donor_models as d
```
nairraghav, line 12: to make this a bit more readable, you may want to consider importing like:
from donor_models import Donor, DonorCollection

This way you can drop the d from before (just a minor cosmetic thing)

students/JoeNunnelley/session09/donor_models.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def donations_to_str(self):
+        """Function to return the donations of this donor as a string"""
+        return ", ".join(str(donation) for donation in self.__donations)
+
+
+class DonorCollection:
+    """The DonorCollection class which contains a set of donors"""
+    def __init__(self, donors=None):
+        self.__donors = []
```
nairraghav, line 61: This is perfectly doable with a list but I would recommend a dict since you need to iterate through the list almost every time you interact with a donor because you have to first find the donor.

students/JoeNunnelley/session09/donor_models.py
```diff
@@ -0,0 +1,121 @@
...
+- Donation
+- DonorCollection
+"""
+
+import datetime
+
+class Donor:
+    """The Donor Class that contains donation information"""
+    def __init__(self, name, _donations=None):
+        self.__donations = _donations
```
nairraghav, line 13: Since you are using the Donation class to track Donations, you may want to do some sort of error checking here otherwise you could add varying types

nairraghav:
Looks good overall Joe, nice work!

PR #179, opened 2019-06-11
lesson9

pberhe:
mailroomOO
test_mailroomOO


students/PeterB/CompletedLesson8/circles.py
```diff
@@ -0,0 +1,62 @@
...
+    def __add__(self, other):
+        return Circle(self.radius + other.radius)
+
+    def __mul__(self, other):
+        return self.radius * other
+
+
+
+
+    @staticmethod
```
nairraghav, line 44: Technically, since you have the __lt__ method created, you do not need to create a method as a sort key. It will use the __lt__ to determine order.

nairraghav:
Nice job with the Circle and previous Mailroom iterations Peter! It looks like you have a good start to the donor_models and tests for Mailroom OO. Don't forget to create a cli_main that actually uses this for the mailroom assignment. Please submit the final Mailroom OO in a future pull request.

PR #180, opened 2019-06-12
mailroom file with classes and test file

colephalen:


students/ColeP/Session09/mailroom_oo.py
```diff
@@ -0,0 +1,185 @@
...
+#         else:
+#             switch[choice]()
+donors_old = {"Donny Donor": [100, 10, 45],
+          "Gav Giver": [12, 16, 20],
+          "Stingy Steve": [2, 5, 1],
+          "Freaky Frank": [1200, 999, 1005],
+          }
+
+
+class Donor:
```
nairraghav, line 29: Generally, we want to pull out our models into their own files

students/ColeP/Session09/mailroom_oo.py
```diff
@@ -0,0 +1,185 @@
...
+        Thank you for your most recent donation of ${:.2f} to The Org!
+
+        -The Org'''.format(self.name, self.donations[-1]))
+
+    @staticmethod
+    def normalize_name(name):
+        return name.lower().strip().replace(' ', '')
+
+
+class StatsTable:
```
nairraghav, line 68: I'm a little skeptical that you need to create this class just for the stats, you should be able to do everything within the donor class itself

students/ColeP/Session09/mailroom_oo.py
```diff
@@ -0,0 +1,185 @@
...
+
+# class MakeDict:
+#     def big_dict(*args):
+#         d = {}
+#         for i in args:
+#             d = {**i}
+#         return d
+
+
+def main_menu():
```
nairraghav, line 108: Don't forget to actually implement your OO changes in this version of mailroom! All your methods should be using the donor models instead of this default donors dict.

nairraghav:
I've given credit for the dict lab. 

It looks you have a good start on the OO mailroom assignment, don't forget to actually use the donor models in your mailroom! Please submit those changes in a new PR

colephalen:
Hi Ron,

Couple questions. What do you mean "actually use the donor models in your
mailroom"? I cleaned up the file a bit (sorry i submitted work with all
those comments) and will resubmit. I also changed main_menu so that my
dictionary is built using class methods.

I'll also try to incorporate the functionality of my second stats class
into the donor class.

Some grading q's:
Can you comment on why I haven't received credit for the fibonacci sequence
exercise and mailroom, pt2? I submitted fib.py a while ago and when working
on a later version of mailroom i incorporated file writing and dictionary
use under the impression that i would receive credit for mailroom 2.

If I complete half of the HTML renderer assignment will i receive half
credit?

Thanks!
Cole

nairraghav:
Hey Cole,

The main goal of the Mailroom OO assignment is to take advantage of the Donor / DonorCollection classes that you created. These classes should handle the logic of the donors / donorcollection. For example, when adding a new donor to mailroom:
1. Use the donor __init__ to create a Donor object
2. Use the donor_collection to insert this Donor to keep track of all donors that exist

The reason you did not receive credit for fib.py was because you did not implement the functions "lucas" and and the "sum_series". I can give some partial credit so that you have something in place for now. 

I never saw a mailroom2 file so i had't given credit but you make a good point. You have implemented the changes involved in it when doing Mailroom3. I will go ahead and give you credit for that!

Yes! You will receive credit for HTMLRender for what you've done. Please let me know when you feel that you have the final submission for it and I will give the appropriate credit for it.

Thanks,
Ron

PR #181, opened 2019-06-13
Adding oop mailroom just in case I don't get back in time from trip

gjkim44:


students/GKim/lesson09/donor_models.py
```diff
@@ -0,0 +1,158 @@
...
+        return key in self.donor_data
+
+    def add_donation(self, key, donation):
+        self.donor_data[key].add_donation(donation)
+
+    def find_donor(self, name):
+        return self.donor_data.get(Donor.key(name))
+
+
+    def find_or_create_donor(self, name, donation):
```
nairraghav, line 93: This would be a good place to make donation=None. When you are looking for a user, you don't need to specify a donation so this a good place for that. You will only need to specify a donation IF you are creating.

nairraghav:
This is great George! Mailroom seems to be functional 👍 
It looks like you may need to implement a couple more tests here, please go ahead with those in a new PR. I will give majority of credit here as a placeholder. Nice work!

PR #182, opened 2019-06-13
Updated series_fibonacci_lucas.py & mailroom3.py

RimleeTalukdar44:
1. Modified series function as instructed below-
Fibonacci: You have the handling of the fibonacci/lucas series correct but the sum series does not work as expected. It calls the lucas when expected but it does not handle the other case. It should handle any starting n0 and n1. In your case, it always assumes n0 and n1 are 1. You should be able to handle any series that start with any n0 and n1. 

2. Modified mailroom3.py as instructed below-
Mailroom 3: Please review your except statements here. When you use "except:", without specifying this exception, this is generally bad behavior because you may accidentally allow an error that shouldn't be allowed. You should specify the error type to be safe.

### DIFF
--- a/students/RimleeT/lesson02/series_fibonacci_lucas.py
+++ b/students/RimleeT/lesson02/series_fibonacci_lucas.py
@@ -21,14 +21,12 @@ Calling this function with no optional parameters will produce numbers from the 
 Calling it with the optional arguments 2 and 1 will produce values from the lucas numbers.
 """
 def sum_series(n, n0=0, n1=1):
-	if n0==2 and n1==1:
-		print("Calls lucas")
-		x=lucas(n)
-		return x
+	if n == 0:
+		return n0
+	elif n == 1:
+		return n1
 	else:
-		print("Calls fibonacci")
-		y=fibonacci(n)
-		return y
+		return sum_series(n - 1, n0, n1) + sum_series(n - 2, n0, n1)
 
 print(fibonacci(5))
 print(sum_series(5))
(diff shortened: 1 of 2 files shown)


students/RimleeT/lesson02/series_fibonacci_lucas.py
```diff
@@ -21,14 +21,12 @@ def lucas(n):
 Calling it with the optional arguments 2 and 1 will produce values from the lucas numbers.
 """
 def sum_series(n, n0=0, n1=1):
-	if n0==2 and n1==1:
-		print("Calls lucas")
-		x=lucas(n)
-		return x
+	if n == 0:
```
nairraghav, line 24: Nice!

students/RimleeT/lesson02/series_fibonacci_lucas.py
```diff
@@ -21,14 +21,12 @@ def lucas(n):
...
+		return n0
+	elif n == 1:
+		return n1
 	else:
-		print("Calls fibonacci")
-		y=fibonacci(n)
-		return y
+		return sum_series(n - 1, n0, n1) + sum_series(n - 2, n0, n1)
```
nairraghav, line 29: 👍

students/RimleeT/lesson05/mailroom3.py
```diff
@@ -44,7 +44,7 @@ def do_thankyou():
 def get_donation_amount():
     try:
         return float(input("Enter the donation amount: "))
-    except:
```
nairraghav, line 47: Nice work!

nairraghav:
Nice job Rimlee, I will go ahead and update grades for these!

PR #183, opened 2019-06-14
Catch up PR

miotke:
Sorry for all the catch up code in this PR.

This contains a fixed html_renderer and session09 code for mailroom_00

Thanks for a great first class Chris, Ron, and Bryan 👏 👍


students/AndrewMiotke/session09/mailroom_oo/mailroom.py
```diff
@@ -0,0 +1,153 @@
...
+    report_rows.sort(key=sort_key, reverse=True)
+
+    print("{:25s} | {:11s} | {:9s} | {:12s}".format("Donor Name", "Total Given", "Num Gifts", "Average Gift"))
+    print("-" * 66)
+
+    for row in report_rows:
+        print("{:25s}   {:11.2f}   {:9d}   {:12.2f}".format(*row))
+
+
+def print_donors():
```
nairraghav, line 36: This is something that could be implemented within the DonorList class itself. Since it has access to all its donors, it could handle the printing instead of putting the logic in your mailroom. 

In general, the logic for the donors should go into that file while logic for the donor_list should go into its class.

Good use of the models but this is something to keep in mind for future OOP! The goal is to put the logic/functions of a specific class into its own class. 

Nice work overall!

nairraghav:
Nice work on the HTML Render and OO mailroom Andrew, please take a look at my comments for future OOP coding! I have also given credit for mailroom 3 since you have applied exception handling in this current version of mailroom.

It was a pleasure having you in the class, looking forward to seeing you in the next one!

miotke:
Thank you Ron, I look forward to reviewing the comments and trying to build on top of that. Thank you for all of your help and I hope we cross paths again.

PR #184, opened 2019-06-14
resubmitting mailroom with one class and initial dictionary built...

colephalen:
... using said class


nairraghav:
Hi Cole,

I have given partial credit for the HTML render work that you have done so far. The OO Mailroom looks good, it would have been nice to implement the DonorCollection class as well (which would handle methods like printing all the donors / generating stats / etc). Overall, you did implement the Donor class and used it to create a dict and other functions so I will give most of the credit for that.

PR #185, opened 2019-06-14
assignment includes Session 6 - 9

kanmenlaohan:
I tried my best on html_render
and OO on mail room I was trying to learn csv module for a db record instead of load data inside code
circle.py should be good to go

### DIFF
--- a/students/Lincoln_Zhang/Session7/test_html_output2.html
+++ b/students/Lincoln_Zhang/Session7/test_html_output2.html
@@ -1,0 +1,2 @@
+None<html_render.Body object at 0x7f0938b8d160>
+</html>
--- a/students/Lincoln_Zhang/Session9/cli_main.py
+++ b/students/Lincoln_Zhang/Session9/cli_main.py
@@ -1,0 +1,66 @@
+#!/usr/bin/env python3
+
+import sys,os,textwrap,csv
+from donor_models import donor_db_collection
+from donor_models import donor
+
+
+
+def donation_accept():
+    donor_name = input("Please tell me your name: ")
+    donor_name = donor_name.capitalize()
+
+    try:
+        donation_amount =list(int(input("Please tell me how much you would like to donate: ")))
+
+
+    except ValueError:
+        print("Please input a number here!")
+
+    with open("donors.csv","a",newline='') as donor_db:
+        donor_db_writer = csv.writer(donor_db)
+        donor_db_writer.writerow([donor_name,donation_amount])
+
+    donors.update({donor_name:donation_amount})
+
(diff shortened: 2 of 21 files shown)


students/Lincoln_Zhang/Session7/test_html_output2.html
```diff
@@ -0,0 +1,2 @@
+None<html_render.Body object at 0x7f0938b8d160>
```
nairraghav, line 1: It looks like you are not printing out the contents of the Element, instead it is printing out the str version of the object.

students/Lincoln_Zhang/Session9/cli_main.py
```diff
@@ -0,0 +1,66 @@
...
+from donor_models import donor
+
+
+
+def donation_accept():
+    donor_name = input("Please tell me your name: ")
+    donor_name = donor_name.capitalize()
+
+    try:
+        donation_amount =list(int(input("Please tell me how much you would like to donate: ")))
```
nairraghav, line 14: you may want to try [(int(input("Please tell me how much you would like to donate: "))] instead, otherwise this throws an error

students/Lincoln_Zhang/Session9/cli_main.py
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env python3
```
nairraghav, line 1: It looks like you don't support adding to a donor that already exists yet, instead it overwrites users when trying to add an existing donor.

nairraghav:
Hi Lincoln,

I have given credit for circle, please take alook at my comments for mailroom. This is pretty cool what you've done but try to complete the add existing donor option. If you would like me to give credit for this as your final submission, let me know and I will give you credit accordingly to completion.

Thanks!

students/Lincoln_Zhang/Session7/test_html_output2.html
nairraghav: I can give some partial credit here since it looks like you have some logic but you are missing some implementation in the render

students/Lincoln_Zhang/Session9/donor_models.py
```diff
@@ -0,0 +1,98 @@
+#!/usr/bin/env python3
+
+import csv, textwrap
+
```
nairraghav, line 4: Generally, you will want to capitalize the first character of your Class names.

nairraghav:
Hi Lincoln,

I will give partial credit for the HTML render since it has some functional parts and some partial credit for the Mailroom OOP as well. It seems to have most of the work in the DonorModels classes but is missing some implementation in cli_main and there are no tests.

PR #186, opened 2019-06-14
Mailroom OO assignment

archwu:
Hi Ron,

Here is my work done on this final assignment, I am still a little confused that how much my work is object-oriented and the meaning of using a module on this project. I was doing unit tests when all the code was in the same file mailroom_oo.py and they were testing well. However, they are not working after I split them. Let me know if there is a good way to unit test the whole module rather than a single file.

Thanks,
Arch

### DIFF
--- a/students/ArchWu/Session8/mailroom_oo.py
+++ b/students/ArchWu/Session8/mailroom_oo.py
@@ -23,7 +23,6 @@ class Donor:
             except:
                 self.donations = [donations]
     
-    
     @property
     def total(self):
         if self.donations:
@@ -85,7 +84,6 @@ class DonorCollection():
     def send_letter():
         pass
     
-
 def main():
     
     prompt = dedent('''
@@ -100,7 +98,7 @@ def main():
     switch = {
         "1": submenu,
         "2": db.report,
-        "3": db.send_letter,
+        "3": save_letters_to_disk,
         "4": goodbye
     }
     
@@ -138,22 +136,6 @@ def goodbye():
 def return_menu():
     return True
 
-# def thank():
-#     while True:
-#         name = input("Enter a donor name (new or existing), type quit to quit: \n >")
-#         if name == 'quit': return
-#         while True:
-#             amount = input("How much have you just donated? > ")
-#             if amount == 'quit': return
-#             try:
-#                 num_donation = float(amount)
-#             except:
-#                 print("Please type a float number")
-#             if Donor.normalize_name(name) in db.donors:
-#                 db.add_donation(name, num_donation)
-#             else:
-#                 db.add_donor(name)
-#                 db.add_donation(name, num_donation)
 
 def thank_you():
     """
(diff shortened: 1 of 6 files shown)


students/ArchWu/Session8/mailroom_oo.py
```diff
@@ -0,0 +1,192 @@
...
+
+def init_donor_db():
+    db = DonorCollection()
+    db.add_donor(Donor("William Gates III", [653772.32, 12.17]))
+    db.add_donor(Donor("Jeff Bezos", [877.33]))
+    db.add_donor(Donor("Paul Allen", [663.23, 43.87, 1.32]))
+    db.add_donor(Donor("Mark Zuckerberg", [1663.23, 4300.87, 10432.0]))
+    return db
+
+class Donor:
```
nairraghav, line 13: You already have this called in your donor models, you should import it instead.

students/ArchWu/Session8/mailroom_oo.py
```diff
@@ -0,0 +1,192 @@
+#!/usr/bin/env python3
+import sys, os, math
+from textwrap import dedent
+
```
nairraghav, line 4: from donor_model import Donor, DonorCollection

students/ArchWu/Session8/mailroom_oo.py
```diff
@@ -0,0 +1,192 @@
...
+    for donor in db.donors.items():
+        with open('{}/{}.txt'.format(temp_path, donor[1].name), 'w') as f:
+            num_donation = donor[1].total
+            greetings = 'Dear {},\n'.format(donor[1].name)
+            body = '\nThank you for your generous gift ${} to us!\n'.format(num_donation)
+            ending = '\nSincerely,\n  ABC foundations'
+            f.write(greetings + body + ending)
+    return
+
+def normalize_name(name):
```
nairraghav, line 186: You already have this in yourDonor object, you can call it from there (Donor.normalize_name)

students/ArchWu/Session8/test_mailroom_oo.py
```diff
@@ -0,0 +1,111 @@
+#!/usr/bin/env python3
+
+"""
+unit tests for the mailroom program
+"""
+import donor_model
```
nairraghav, line 6: if this is import in your mailroom_oo, you will not need to call it here.

nairraghav:
Hey Arch, 

You should be able to have it functional when you import the donor_model file in your mailroom_oo. When you import that and run your tests, what do you see?

archwu:
> Hey Arch,
> 
> You should be able to have it functional when you import the donor_model file in your mailroom_oo. When you import that and run your tests, what do you see?

It was like no test was run and nothing happens

students/ArchWu/Session8/mailroom_oo.py
archwu: Actually this mailroom_oo is the file I was working on before I saw the requirement to split into multiple files and it happen to have the same file name as the requirement mentioned. What I did try is simply cutting those class initialization code into donor_model and split into two files.

nairraghav:
Hey Arch, I have copied over your files, removed the classes into their own files called "donor_models" and ran the test. It worked successfully for me
Screen Shot 2019-06-15 at 10 59 17 AM

nairraghav:
I will give you credit for what you have since you pretty much have it working ✅

PR #188, opened 2019-06-15
'mailboxCLI'

pberhe:
lesson 9 mailbox CLI

### DIFF
--- a/students/PeterB/Lesson09/CLI.py
+++ b/students/PeterB/Lesson09/CLI.py
@@ -1,0 +1,60 @@
+#!/usr/bin/env python3
+
+import sys
+
+DONOR_DICT = d.DonorDB()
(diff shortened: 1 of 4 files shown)


students/PeterB/Lesson09/CLI.py
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env python3
+
+import sys
+
```
nairraghav, line 4: Don't forget to import your donor_models file

nairraghav:
As per your commit message, I have given some small partial credit for the work you have done so far in HTML render. As it is mostly incomplete, I don't think I would be able to give more. Please try to complete the mailroom OO  and any other missing assignments in  a future pull request. Let me know if you have any questions or if I can help!

Thanks,
Ron

PR #191, opened 2019-06-15
'updatedimports'

pberhe:
updatedCLIimports

### DIFF
--- a/students/PeterB/Lesson09/CLI.py
+++ b/students/PeterB/Lesson09/CLI.py
@@ -1,6 +1,9 @@
 #!/usr/bin/env python3
 
 import sys
+from Lesson9mailboxmoneyOO import Donor
+from Lesson9mailboxmoneyOO import DonorDB
+
 
 DONOR_DICT = d.DonorDB()
 DONOR_DICT.add(d.Donor("Road Runner", [4839283.32, 17842102.94]))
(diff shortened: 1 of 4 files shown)


students/PeterB/Lesson09/CLI.py
```diff
@@ -1,6 +1,9 @@
 #!/usr/bin/env python3
 
 import sys
+from Lesson9mailboxmoneyOO import Donor
```
nairraghav, line 4: Since you're importing as Donor and DonorDB, you dont need to prepend "d" before classing them

students/PeterB/Lesson09/CLI.py
```diff
@@ -1,6 +1,9 @@
...
+from Lesson9mailboxmoneyOO import DonorDB
+
 
 DONOR_DICT = d.DonorDB()
```
nairraghav, line 8: This would be: DONOR_DICT = DonorDB()

nairraghav:
It looks like your mailroom doesn't have all functionality and no tests, I can give some partial credit. I've added credit for the python pushups in this PR and you already had full credit for the slicing lab. Let me know if you have any other questions!

pberhe:
that's odd that my test was not apart of the pull. i have resubmitted my
test for the mailroom assignment

PR #193, opened 2019-06-16
Please review and grade the mailroom oop

RimleeTalukdar44:
1. Added cli_main.py
2. Added donor_models.py
3. Deleted old mailroom file which I added by mistake.

I could not get time to get test cases for mailroom oop.


students/RimleeT/lesson09/mailroom_oo/donor_models.py
```diff
@@ -0,0 +1,81 @@
...
+        letter = f"Dear {self.name}\n\n\tThank you for your very kind donation of ${stat['Total']} It will be put to very good use.\n\nSincerely,\nThe Team"
+        
+        f = tempfile.NamedTemporaryFile(prefix = self.name + '-', dir = dir_name, delete = False)
+        f.write(letter.encode())
+        
+        return letter
+        
+class DonorCollection():
+    def __init__(self):
+        self._donor_list = list()
```
nairraghav, line 50: It may be a bit better to use a Dict instead of an array because any time you want to return a Donor's information, you have to iterate through the list to find the Donor. Instead, if you were to use a Dict, you could map it to the name (since they should be unique) and it would be more performant!

nairraghav:
Excellent job Rimlee! I saw that you did not have time for tests but the functionality looks great ✅

